PR #4, opened 2026-04-05
Add --render and --upload options to fit_model.py

frankbuckley:
## Summary
- Add `--render` flag to invoke Quarto rendering after model fitting
- Add `--upload` flag to sync model output to Azure Blob Storage via AzCopy (reads container URL from `DSERESEARCH_BLOB_CONTAINER_URL` env var); each upload is placed under a unique UUIDv7 (timestamp-based) prefix before `model_label` in the destination path
- Refactor model dispatch in `fit_model.py` to correctly support `all` (previously only ran VG01 due to `elif` chain)
- Extract upload logic into `src/vocab_growth/storage.py` module
- Synchronise `CLAUDE.md`, `AGENTS.md`, and `.github/copilot-instructions.md` with updated command docs and add keep-in-sync notice

## Test plan
- [ ] Run `python scripts/fit_model.py vg01 --config dev` — verify basic fitting still works
- [ ] Run with `--render` — verify Quarto is invoked on output
- [ ] Run with `--upload` without `DSERESEARCH_BLOB_CONTAINER_URL` set — verify clear error message
- [ ] Run with `--upload` and env var set — verify AzCopy invocation with correct source/destination, including UUIDv7 prefix in destination path
- [ ] Run `ruff check scripts/fit_model.py src/vocab_growth/storage.py` — no lint errors

🤖 Generated with <a href="https://claude.ai/claude-code">Claude Code</a>

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,5 +1,7 @@
 # Copilot Instructions
 
+> **Keep in sync:** This file, `CLAUDE.md`, and `AGENTS.md` share the same content. When updating one, update all three.
+
 ## Project overview
 
 This project is an exploratory study of vocabulary development in children with Down syndrome that aims to characterise observed trajectories of word learning, spoken and gestured production, and relationships between words understood and produced. The primary goal of the study is to provide interpretable statistics that can accurately inform expectations, intervention and teaching practice. We evaluate and fit these models using Bayesian inference to estimate full probability distributions for parameters of interest, using an iterative workflow.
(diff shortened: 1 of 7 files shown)


frankbuckley:
@copilot -- I think the location the output is uploaded to in storage.py should include a timestamp - please update to include a version 7 guid before model_label.

frankbuckley:
@copilot -- why does every commit cause CI workflow twice - for pull-request and push -- please fix.

PR #43, opened 2026-06-09
feat(models): add VG11, VG12, VG13 — TD models with study random intercepts

ethanbuckley:
## Summary

- **VG11** — Words spoken (TD): extends VG03 with dataset-level study random intercepts (non-centred) and a GP anchor at 19 months to remove the GP–intercept ridge
- **VG12** — Words understood (TD): extends VG04 with the same study RE structure (WG + Oxford CDI only; WS excluded)
- **VG13** — Bivariate understood + spoken (TD): extends VG06 but **restricted to ages 8–18 months** where WG/Oxford CDI coverage is dense across multiple labs; study REs on both trajectories; GP anchors at 13 months

VG13 addresses the data coverage gap above 18 months in VG06 (after the WS fix, only a single dataset contributes bivariate observations above that age, making inference unreliable).

## Infrastructure changes

- `data_utils.load_data()`: adds `dataset_name as study` alias for TD rows (so all RE pipelines can use a unified `"study"` column regardless of population), and adds a `max_age_months` parameter
- `definitions.BivariateModelDefinition`: new `max_age_months` field, threaded into `common_bivariate_re.prepare_bivariate_re_data`
- `common_univariate_re.py`: new shared pipeline for univariate RE models (data prep, model build, fit orchestration)
- `MODEL_REGISTRY`, `fit_model.py`, `upload.py`: vg11/vg12/vg13 registered; `python scripts/fit_model.py vg11` etc. work immediately
- Quarto docs for vg11/vg12/vg13, updated CLAUDE.md/AGENTS.md/copilot-instructions.md

## Test plan

- [ ] `ruff check src/ scripts/` — passes clean ✅
- [ ] `npm run spellcheck` — passes clean ✅
- [ ] `npm run format:check` — passes clean ✅
- [ ] Smoke-test each model with `python scripts/fit_model.py vgNN --config dev` once the database is available
- [ ] Check VG11 prior predictive trajectories look reasonable (spoken 9–30 months, TD)
- [ ] Check VG12 prior predictive trajectories look reasonable (understood 9–30 months, TD)
- [ ] Check VG13 prior predictive trajectories look reasonable (bivariate 8–18 months, TD)


ethanbuckley:
All local linter, formatting, and spellcheck tests pass cleanly. This PR establishes the core univariate random effects pipeline (`common_univariate_re.py`), migrates the TD query logic to support dataset-level grouping, and implements `VG11`, `VG12`, and the age-capped `VG13` early milestone model. Ready for review and merge.

src/vocab_growth/data_utils.py
```diff
@@ -98,12 +104,17 @@ def load_data(
...
+    age_upper = max_age_months if max_age_months is not None else 30
+
     with duckdb.connect(VOCABULARY_DATA_PATH) as con:
         td_df = (
             con.execute(
                 """
             SELECT
                 form,
+                dataset_name,
+                dataset_name                       as study,
```
frankbuckley: The plain `dataset_name` column is redundant here — no caller requests it (the RE pipelines select `"study"`), and `td_df[columns]` drops anything unrequested. Keeping just the aliased form is enough:

```suggestion
                dataset_name                       as study,
```

src/vocab_growth/models/definitions.py
```diff
@@ -410,6 +427,94 @@ def model_type(self) -> ModelType:
...
+        "with dataset-level study random intercepts"
+    ),
+    population=Population.TYPICALLY_DEVELOPING,
+    # Common 800-item reference inventory. Counts from WG (ceiling 396) and
+    # Oxford CDI (ceiling 418) are comfortably below the 800-item scale in the
+    # 8–18 month window, so no distortion is introduced.
+    n_trials=800,
+    # Restrict to 8–18 months where WG/Oxford CDI data are dense and the WS
+    # bias (production proxy comprehension) is avoided entirely.
+    max_age_months=18,
```
frankbuckley, line 495: `max_age_months=18` enforces only the **upper** bound of the documented “8–18 months” window. There's no lower-bound filter in `load_data` or `prepare_bivariate_re_data`, so the 8-month floor relies entirely on the natural minimum age of the WG/Oxford CDI data. If any rows fall below 8 months they'll be silently included, contradicting the window described in the docstring and the `n_trials=800` ceiling rationale.

Could you confirm the data minimum for these forms is ≥ 8 months? If not, consider adding a `min_age_months` filter (symmetric with `max_age_months`) so the window is enforced in code rather than assumed.

frankbuckley (commented):
Re-reviewed at `47079839e`. This is a clean reuse of the existing fit pipeline and the VG10 GP-anchor pattern, and the data-infra changes are backward-compatible for DS. One blocking item plus a few smaller points for your consideration — details inline.

### 🔴 Blocking — breaks `tests/test_data_utils.py`

The TD branch of `load_data` now `SELECT`s `dataset_name` and `child_id`, but the fixture `_create_wordbank_db` in `tests/test_data_utils.py` (not touched by this PR) creates `wordbank_child` without those columns. All three TD tests (`test_td_load_data_keeps_ws_as_spoken_only`, `test_td_understood_data_excludes_ws_before_sampling`, `test_td_spoken_data_includes_ws_and_bivariate_forms`) fail at bind time:

```
BinderException: Referenced column "dataset_name" not found in FROM clause!
```

CI stays green because `ci.yml` runs only spellcheck/ruff and a single `vg01` (DS) fit — it never runs `pytest` or any TD model. (Note `CLAUDE.md` documents `pytest` as the test command.) The file isn't in the diff so I can't attach an inline suggestion; the fix is to add the two columns to the fixture and the `INSERT` rows:

```python
        con.execute(
            """
            CREATE TABLE wordbank_child (
                form VARCHAR,
                dataset_name VARCHAR,
                child_id VARCHAR,
                age DOUBLE,
                comprehension INTEGER,
                production INTEGER,
                typically_developing BOOLEAN,
                health_conditions VARCHAR
            )
            """
        )
        con.executemany(
            """
            INSERT INTO wordbank_child VALUES (?, ?, ?, ?, ?, ?, ?, ?)
            """,
            [
                ("WG", "Fenson (2007)", "c01", 12.0, 40, 12, True, None),
                ("Oxford CDI", "Hamilton (2000)", "c02", 24.0, 190, 80, True, None),
                ("WS", "Fenson (2007)", "c03", 28.0, 51, 51, True, None),
                ("WSShort", "Fenson (2007)", "c04", 22.0, 20, 20, True, None),
                ("TEDS Twos", "Fenson (2007)", "c05", 24.0, 75, 75, True, None),
                ("WG", "Fenson (2007)", "c06", 35.0, 100, 50, True, None),
                ("WG", "Fenson (2007)", "c07", 18.0, 60, 20, True, "premature"),
                ("WG", "Fenson (2007)", "c08", 18.0, 60, 20, False, None),
            ],
        )
```

A small regression test asserting `study` (and `subject_id` when requested) come back populated would also lock in the new aliasing behaviour.

docs/models/vg13/index.qmd
```diff
@@ -0,0 +1,108 @@
...
+
+```{python}
+import matplotlib.pyplot as plt
+import pandas as pd
+```
+
+## Statistical model
+
+We model words understood and words spoken as bounded count outcomes out of an 800-word reference inventory, restricted to children aged 8–18 months. The understood trajectory and the production ratio $q = p_S / p_U$ are each modelled as smooth functions of age (linear trend + HSGP), with each GP anchored at 13 months to remove the GP–intercept ridge. Dataset-level study random intercepts (non-centred Normal) are added to both trajectories, absorbed at the observation level only. Plot and query predictions use the population-level trajectory. A Beta-Binomial likelihood with age-varying dispersion accounts for extra-binomial variability in both outcomes.
```
frankbuckley: VG13 routes through `common_bivariate_re.build_model_re`, where the study intercepts use a **centred** parameterisation (`delta_u = pm.Normal("delta_u", mu=0, sigma=tau_u)`), not non-centred. Only the new univariate pipeline (VG11/VG12) is genuinely non-centred. Suggest correcting the description:

```suggestion
We model words understood and words spoken as bounded count outcomes out of an 800-word reference inventory, restricted to children aged 8–18 months. The understood trajectory and the production ratio $q = p_S / p_U$ are each modelled as smooth functions of age (linear trend + HSGP), with each GP anchored at 13 months to remove the GP–intercept ridge. Dataset-level study random intercepts (centred Normal) are added to both trajectories, absorbed at the observation level only. Plot and query predictions use the population-level trajectory. A Beta-Binomial likelihood with age-varying dispersion accounts for extra-binomial variability in both outcomes.
```

Separately (out of scope for a doc fix): the centred form can funnel with few studies under `HalfNormal(0.5)` on `tau`, so you might later consider matching VG11/VG12's non-centred study REs in `common_bivariate_re`.

src/vocab_growth/models/common_univariate_re.py
```diff
@@ -0,0 +1,512 @@
...
+        # Population-level linear predictor (no study effect)
+        f_all = pm.Deterministic("f_all", mean_trend_all + g, dims=("all_id",))
+
+        # ============================================================
+        # Study-level random intercepts (non-centred parameterisation)
+        # ============================================================
+
+        tau = pm.HalfNormal("tau", sigma=definition.tau_study_sigma)
+        delta_raw = pm.Normal("delta_raw", mu=0.0, sigma=1.0, dims="study_id")
+        delta = pm.Deterministic("delta", tau * delta_raw, dims="study_id")
```
frankbuckley, line 340: These study-level random intercepts are added unconditionally — `build_univariate_re_model` never reads `definition.use_study_re`. That makes the flag (set `True` on VG11/VG12) effectively dead: a future univariate model routed through this pipeline with `use_study_re=False` would still get study REs. Consider either honouring the flag here (gate this block on `definition.use_study_re`, with the obs-level shift falling back to `0.0` when off) or dropping the field so it doesn't read as wired-up.

src/vocab_growth/data_utils.py
ethanbuckley: Removed, keeping only dataset_name as study

src/vocab_growth/models/definitions.py
ethanbuckley: WG minimum age in the data is 8 months, Oxford CDI minimum is 12. No rows fall below the 8-month floor so no code change needed.

docs/models/vg13/index.qmd
ethanbuckley: Fixed — changed to 'centred Normal'

src/vocab_growth/models/common_univariate_re.py
ethanbuckley: Removed the field entirely from the dataclass and both VG11/VG12 definitions. Study REs are always on for univariate RE models.

Issue #49, opened 2026-06-15
Model signing: proportion understood signed by age, and link to spoken (VG14 → VG15)


frankbuckley:
## Summary

Extend the joint understood/spoken models to incorporate **signing** as a third production modality for children with Down syndrome. The goal is two-fold:

1. Characterise the **proportion of understood words that are signed** as a function of age.
2. Model the **relationship between signing and spoken production**

This issue proposes a **staged** delivery: a lightweight model (Option 1) that reuses the existing bivariate machinery and runs on the full DS sample, followed by a fuller joint model (Option 3) that estimates the within-understood association between signing and speaking. Option 1 should be built first; its posteriors provide sensible priors for Option 3.

## Background

The bivariate models (VG05–VG10, see `src/vocab_growth/models/common_bivariate.py`) use a production-ratio reparameterisation:

- `p_U(a) = sigmoid(f_U(a))` — proportion of the checklist understood
- `q(a)   = sigmoid(h(a))`   — of understood words, the fraction **spoken**
- `p_S    = p_U · q`         — guarantees `p_S ≤ p_U` by construction

Signing extends this naturally with a parallel ratio.

### Data availability

Signing appears only in the **Down syndrome** datasets (the TD / Wordbank data has none), so this is a DS model in the VG05/VG07–VG10 family. `signed` already flows through the `vocab_combined` view (selected in `scripts/prepare_data.py`) but no model consumes it yet.

| Dataset | n | signing field(s) | ages (months) |
| ------- | --- | --------------------------------------------- | ------------- |
| uk_01   | 219 | `signed` (marginal; 63 > 0)                   | 15–115        |
| uk_02   | 136 | `signed_only`, `signed_spoken`, `signed`      | 19–56         |
| uk_04   | 44  | `signed`                                      | 18–45         |
| uk_05   | 48  | `signed`                                      | 17–36         |
| uk_06   | 11  | `signed`                                      | 60–115        |

**Key subtlety:** signing and speaking are **not mutually exclusive**. uk_02 records `signed_only`, `spoken_only` and `signed_spoken` separately — a child may both sign and speak the same word. So within understood words there is a 2×2 structure (signed yes/no × spoken yes/no) → four cells: *neither / sign-only / speak-only / both*. Only **uk_02** records all four cells; the other datasets give only the marginal `signed` and `spoken` totals (`signed = signed_only + both`, `spoken = spoken_only + both`).

---

## Option 1 — Parallel ratios (deliver first)

[...]

PR #52, opened 2026-06-15
Add VG14 — signing as a third production modality (trivariate model)

ethanbuckley:
## Summary

Implements **Option 1 of #49**: a trivariate model of words understood, spoken, and signed for children with Down syndrome. It extends the bivariate production-ratio reparameterisation with a third ratio for signing, hanging off the same understood trajectory:

- `p_U(a) = sigmoid(f_U(a))` — proportion understood
- `q(a) = sigmoid(h(a))`, `p_S = p_U·q` — spoken (unchanged)
- `r(a) = sigmoid(g(a))`, `p_Sign = p_U·r` — signed (new), enforces `p_Sign ≤ p_U` by construction
- third Beta-Binomial likelihood `y_sign_obs` with its own age-varying `kappa_sign`
- derived `p_any = p_U·[1−(1−r)(1−q)]` — total expressive vocabulary

Built in a new isolated engine `common_trivariate.py` (a copy-and-extend of `common_bivariate.py`). `common_bivariate.py` / `common_bivariate_re.py` / VG01–VG13 are **untouched** — isolation was the explicit design choice.

## Headline result (rep fit: 6 chains × 6000 draws)

- Clean sampling: **0 divergences / 36,000, max r̂ 1.001, min ESS ≈ 6.5k**.
- The robust finding: signing rises from near-zero in infancy to ~0.4–0.5 of understood words by age 2–3, then hands off to speech — `r(a)` crosses the rising `q(a)` at **~39 months**.
- `p_any` exceeds spoken-alone at every age (e.g. ~135 vs 64 words at 36 mo).

## Caveats deliberately baked into the report/notes (please read before relying on the curves)

1. **`p_any` is an independence-based upper bound.** VG14 only uses marginal signed/spoken totals; the sign∩speech overlap is unobserved except in uk_02. Validated against uk_02's four cells: independence over-states the union by **~3.7 pp** (sign and speech are positively associated). VG15 will estimate it properly.
2. **The signed peak age and recede are study-confounded, not identifiable here.** The 48–60 mo window is entirely uk_02; the apparent recede is the uk_02→uk_01 sample boundary. VG14 has no study random effects (mirrors VG05), so the age curve absorbs study composition. Only the rise (→ age 2–3) and the ~39 mo crossover are reported as robust; the peak month is flagged uncertain.
3. **uk_06 included in the signed likelihood** (`include_uk06=True` default): its `signed` is a real signing-production count (often comparable to or exceeding spoken at older ages), not a comprehension measure — signing implies understanding. The flag is kept for reversibility. Including it leaves the headline unchanged (re-fit clean: 0 divergences/36k, r̂ 1.001) and only lifts the old-age signed tail / weakens the recede. The open question for Frank/Sue is now whether uk_06's signed counts are **coded comparably** to uk_02/04/05 (uk_06 has no field dictionary), not the construct.
4. **One intentional divergence from the bivariate engine:** the full-grid `*_all` intermediates are plain tensors, not stored `pm.Deterministic` nodes (they're never extracted — only obs/plot/query slices are). This fixed an OOM on the 6×6000 rep fit and is worth porting back to the bivariate models.

## Files

- New: `src/vocab_growth/models/common_trivariate.py` (engine), `model_vg14.py` (wrapper), `docs/models/vg14/index.qmd` (report), `notes/202606151500-…` (baseline), `notes/202606151700-…` (review).
- Modified: `definitions.py` (`ModelType.TRIVARIATE`, `TrivariateModelDefinition`, `VG14`, registry), `scripts/fit_model.py` (dispatch), VG14 row in `CLAUDE.md` / `AGENTS.md` / `.github/copilot-instructions.md`, `config/spellcheck/allow-en.txt` (+`toddlerhood`).
- `output/` and `data/*.duckdb` are gitignored — fitted artifacts not tracked.

## Test plan

- [x] `ruff check src/ scripts/` passes
- [x] `npm run spellcheck` — 0 issues
- [x] Prettier `--check` clean
- [x] dev fit clean; rep fit 0 divergences, r̂ 1.001, min ESS ≈ 6.5k
- [x] `docs/models/vg14/index.qmd` renders, no broken figures
- [x] `common_bivariate*.py` and VG01–VG13 unchanged in the diff

## Open questions for reviewers

- uk_06: confirm its signed counts are coded comparably to uk_02/04/05 (now included by default).
- Should a study-RE variant follow (peak-age identifiability), or is that deferred to VG15?
- OK to report `p_any` as an explicit upper bound rather than a point estimate?

## Follow-ups (not in this PR)

- VG15 — uk_02 four-cell multinomial: estimates the overlap (replaces the `p_any` upper bound) and adds study random effects.
- Port the `*_all`-as-plain-tensors memory fix back to the bivariate engine.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -61,7 +61,7 @@ This merges CSV datasets from `data/` into `data/vocab_data_merged.csv` and a Du
 python scripts/fit_model.py <model_id> [--config <config>] [--render] [--upload]
 ```
 
-- `model_id`: one of `vg01`, `vg02`, `vg03`, `vg04`, `vg05`, `vg06`, `vg07`, `vg08`, `vg09`, `vg10`, `vg11`, `vg12`, `vg13`, or `all`.
+- `model_id`: one of `vg01`, `vg02`, `vg03`, `vg04`, `vg05`, `vg06`, `vg07`, `vg08`, `vg09`, `vg10`, `vg11`, `vg12`, `vg13`, `vg14`, or `all`.
 - `--config`: sampling configuration — `dev` (fast, for development), `test`, or `rep` (full reporting quality). Defaults to `dev`.
 - `--render`: render the Quarto model output after fitting.
 - `--upload`: upload model output to Azure Blob Storage via AzCopy. Requires `DSERESEARCH_BLOB_CONTAINER_URL` environment variable set to the target container URL.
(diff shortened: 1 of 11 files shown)


frankbuckley (approved):

ethanbuckley:
### Follow-up `da5750a` — intercept-only signed mean + Copilot review items

**Freed the prior-dominated signed level.** The signed mean-trend anchors (`Beta(15,90)`) were prior-dominated — posterior sd ≈ prior sd, so the signed *level* was set by the prior, not the data (the understood/spoken anchors all moved and shrank, so the strictness was isolated to the signed mean). Fix: the signed mean is now **intercept-only** (the age *slope* was the actual "spurious 58% at 12 mo" failure mode; dropping it removes that structurally), with one weakly-informative intercept `Normal(logit(0.15), 0.75)`.

Re-fit at `--config rep` is clean (**0 divergences / 36k, r̂ 1.001, min ESS ≈ 7.6k**):
- `intercept_sign` posterior **sd 0.44 vs prior 0.75** → now data-informed.
- `eta_sign` 1.15 → 1.06 (not pinned); `ell_sign` not pinned.
- **Curve unchanged**: hump peak ~0.47 @ ~30 mo, crossover ~39 mo, tail r(90) ≈ 0.04.

Honest note: contrary to the hypothesis that uk_06's heavy older signers would lift the old-age tail, the freed level lands at ~0.14 and the tail stays low — uk_01's many near-zero older signers outweigh uk_06's few heavy ones, so the data genuinely supports a low old-age population tail. The peak does **not** relocate and isn't expected to: looser priors give honest wider uncertainty, not a later peak. Peak timing remains study-confounded; the real fix is study random intercepts (VG15), not prior tuning. Details in `notes/202606151700-…` §6.

**Copilot review addressed** (same commit):
- The two stale `common_trivariate.py` comments (uk_06 "excluded by default" / "comprehension-adjacent construct") are corrected to match `include_uk06=True` and the "real signing count" framing.
- The report grammar nit ("more approximate than _where_ …" → "_when_ …") is fixed.

PR #85, opened 2026-06-29
Fix holdout-mask misalignment in bivariate-RE extract_model_samples (#67)

### LINKED ISSUE #67: Holdout-mask misalignment in bivariate-RE extract_model_samples (K-fold LOSO path)
Flagged during the review in #64 (latent; does not affect standard fits).

## Observation
In `src/vocab_growth/models/common_bivariate_re.py` (`build_model_re`), the observation masks are stored from the **full** observed masks:

```python
_ = pm.Data("obs_u_mask", has_u.astype(int), dims=("obs_id",))
_ = pm.Data("obs_s_mask", has_s.astype(int), dims=("obs_id",))
```

but the likelihood is fit on the **training** rows only (`has_u_train = has_u & ~holdout`), so `observed_data["y_u_obs"]` / `["y_s_obs"]` have length `sum(has_u_train)` / `sum(has_s_train)`.

`extract_model_samples` (in `common_bivariate.py`) then does:

```python
y_u_obs = np.full(n_obs, np.nan)
y_u_obs[obs_u_mask] = y_u_obs_raw   # obs_u_mask has sum(has_u) Trues; raw has sum(has_u_train)
```

When a `holdout` column is present, `sum(has_u) > sum(has_u_train)`, so this assignment is **misaligned** (shape mismatch or silently wrong scatter).

## Scope
- **Standard fits are unaffected**: with no `holdout` column, `has_u_train == has_u`, so the masks and `observed_data` line up exactly. All committed VG07–VG13 outputs are fine.
- Only the K-fold LOSO scripts (`scripts/kfold_loso.py`, `scripts/loso_compare*.py`) inject a `holdout` column. They may not currently route through the standard `extract_model_samples` path — this should be confirmed.

## Suggested fix
Store the masks from the **training** masks (`has_u_train` / `has_s_train`) so they align with `observed_data`, or guard the extraction explicitly when a holdout is present.

## Suggested acceptance criteria
- [ ] Confirm whether any LOSO script reaches `extract_model_samples` with a holdout active.
- [ ] Align the stored masks with the training rows (or document why the full masks are intended and guard the scatter).
- [ ] Add a regression test that builds the bivariate-RE model with a synthetic `holdout` column and round-trips `extract_model_samples` without misalignment.

ethanbuckley:
> [!NOTE]

Closes #67.

## Problem

`build_model_re` (`common_bivariate_re.py`) stored the **full** observed masks (`has_u` / `has_s`), but fits the likelihood on training rows only (`has_*_train = has_* & ~holdout`). So `observed_data["y_*_obs"]` has length `sum(has_*_train)`, while the stored `obs_*_mask` has `sum(has_*)` `True`s. When a `holdout` column is present, `sum(has_*) > sum(has_*_train)`, and the scatter in `extract_model_samples` —

    y_u_obs = np.full(n_obs, np.nan)
    y_u_obs[obs_u_mask] = y_u_obs_raw

— is misaligned.

**This is latent on `main`:** standard fits have no `holdout` column (so `has_*_train == has_*`), and the only producer of a `holdout` column, `scripts/kfold_loso.py`, reads the posterior directly and never reaches `extract_model_samples`. Committed VG07–VG13 outputs are unaffected.

## Changes

- **`common_bivariate_re.py`** — `build_model_re` stores the training masks (`has_*_train`) so `obs_*_mask` aligns with the likelihood rows / `observed_data`. No-op when there is no `holdout` column.
- **`common_bivariate.py`** — `extract_model_samples` now raises a clear `ValueError` (referencing #67) if a stored mask's count diverges from its observed-data length. Defends the non-RE path too.
- **`tests/test_bivariate_re_holdout_mask.py`** (new) — builds the real VG07 model with a synthetic `holdout` column and round-trips the genuine extraction path (`sample_posterior_predictive` → `extract_model_samples`); plus a guard test. Verified to fail without the fix.

## Verification

- `pytest` → 41 passed, 2 skipped.
- `ruff check src/ scripts/ tests/` → clean.
- Reverting the mask fix makes the new round-trip test fail with the `#67` guard error; restoring it passes.

### DIFF
--- a/src/vocab_growth/models/common_bivariate.py
+++ b/src/vocab_growth/models/common_bivariate.py
@@ -752,10 +752,22 @@ def extract_model_samples(trace: xr.DataTree) -> BivariateModelSamples:
     n_obs = len(obs_u_mask)
 
     y_u_obs_raw = np.array(trace.observed_data["y_u_obs"].values, dtype=float)
+    if int(obs_u_mask.sum()) != y_u_obs_raw.shape[0]:
+        raise ValueError(
+            f"obs_u_mask count ({int(obs_u_mask.sum())}) does not match observed "
+            f"y_u_obs length ({y_u_obs_raw.shape[0]}); stored mask and likelihood "
+            "rows are misaligned (issue #67)."
+        )
     y_u_obs = np.full(n_obs, np.nan)
     y_u_obs[obs_u_mask] = y_u_obs_raw
 
     y_s_obs_raw = np.array(trace.observed_data["y_s_obs"].values, dtype=float)
+    if int(obs_s_mask.sum()) != y_s_obs_raw.shape[0]:
+        raise ValueError(
+            f"obs_s_mask count ({int(obs_s_mask.sum())}) does not match observed "
+            f"y_s_obs length ({y_s_obs_raw.shape[0]}); stored mask and likelihood "
+            "rows are misaligned (issue #67)."
+        )
     y_s_obs = np.full(n_obs, np.nan)
     y_s_obs[obs_s_mask] = y_s_obs_raw
 
--- a/src/vocab_growth/models/common_bivariate_re.py
+++ b/src/vocab_growth/models/common_bivariate_re.py
@@ -374,9 +374,13 @@ def build_model_re(
         _ = pm.Data("X_plot", X_plot.flatten(), dims=("plot_id",))
         _ = pm.Data("X_query", X_query.flatten(), dims=("query_id",))
 
-        # Store masks and indices as constant data for extraction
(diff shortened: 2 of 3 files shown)


frankbuckley (approved):

frankbuckley (commented):
> [!NOTE]
> Drafted by an LLM-based AI tool (GitHub Copilot/GPT-5.5).

Reviewed the holdout-mask alignment changes carefully, including the bivariate random-effects sample extraction path and surrounding LOSO masking logic. I don't have any line-specific findings to raise; the mask alignment fix looks correct and appropriately guarded.

PR #117, opened 2026-07-03
feat(vg15): wire in New Zealand (nz_01) data set

### LINKED ISSUE #110: Add New Zealand data set
When prepared, wire in New Zealand data set.

Note: no comprehension, only production, but includes sign only, sign+say and say only, which will help with disentangling estimates of total (sign ∪ spoken).

ethanbuckley:
> [!NOTE]

## Summary

Wires the New Zealand (nz_01) dataset in. Closes #110. Depends on `dsegroup/research-data-analysis#8`.

Per #110, nz_01 is production-only but records **say-only / sign-only / both**, which helps disentangle the sign ∪ spoken total. This PR exposes its marginals to the DS models (**B1**) and feeds its produced cross-tab into VG15's sign–speech association `psi` (**B2**).

## B1 — data availability (`scripts/prepare_data.py`)

nz_01 enters `vocab_combined` with `understood = NULL` (no comprehension). The modality-exclusive CSV columns map to the any-modality marginals:

- `spoken = spoken + spoken_signed` (a + c)
- `signed = signed + spoken_signed` (b + c)
- `produced = a + b + c`, `survey_vocab_max = 675`

Because `load_data(DOWN_SYNDROME)` pools all studies, nz_01 is then available to the marginal DS models (VG01/05/07–10/14) on their next fit.

## B2 — VG15 rich cross-tab (`common_joint_modality.py`, `definitions.py`)

nz_01's `spoken`/`signed`/`spoken_signed` are a produced modality cross-tab, but it **cannot** fill uk_02's *within-understood* four cells — nz_01 has no comprehension, so its "neither" mixes understood-only with not-understood. Instead we **condition on produced**: drop the unobservable `understood_only` cell from the Plackett composition and renormalise over `{sign_only, speak_only, both}`, giving a within-produced **3-cell Dirichlet-Multinomial** (`nz_prod_cells_obs`) that reuses the same `psi` and `conc`. So nz_01 (111 rows across 30–66 mo) informs VG15's `psi`. Excluded from the marginal path (like uk_02's four-cell rows) to avoid double counting; gated by `include_nz01_cells` (default `True`).

## Verification

`python scripts/fit_model.py vg15 --config dev` runs end-to-end:

- `nz_01 produced-cell rows (DM) = 111`; **0 divergences**.
- `psi` median **1.71**, 90% HDI [1.19, 2.37], P(psi > 1) = 0.99.
- (Dev-tier max R̂ ≤ 1.06 / low ESS on the subject REs are 500-draw artifacts.)

## Follow-ups (not in this PR)

- `data/vocab_data_nz_01.csv` (real `DSE_ANONYMISATION_KEY`) must be present/committed for the pipeline to run.
- Marginal DS models score against a fixed `n_trials = 800`, not the 675 ceiling (pre-existing heterogeneity, shared with uk_03=418 etc.); the B2 produced-DM is unaffected (total is the observed produced count).
- Rep-tier re-fit + report regen of VG15 and the marginal DS models; optional `psi` with/without `include_nz01_cells`; extending PPC to `nz_prod_cells_obs`.


ethanbuckley:
> [!NOTE]

Reviewed — statistically sound and carefully guarded. I pushed a unit test (ab2c248) to strengthen it.

- **The produced-conditioned 3-cell DM is correct.** Conditioning on produced = dropping the unobservable "neither" / understood-only cell from the Plackett composition and renormalising over `{sign_only, speak_only, both}` (`pi_sign = r − both`, `pi_speak = q − both`, `pi_both`, then `/ sum`) — exactly `P(cell | produced)`. Cell ordering matches between `prod_counts` and `pi_prod`, and it reuses the **same `psi` / `conc` / population `q, r`** as the uk_02 four-cell, so the two cross-tab sources jointly identify `psi`. Sum/negativity guards are in place.
- **Marginal mapping** (`spoken = a+c`, `signed = b+c`, `produced = a+b+c`, ceiling 675) is right, and double-counting is avoided (`other` excludes both uk_02 and nz_01; nz_01 enters VG15 only via the DM).
- **Missing-CSV handling is graceful** (empty-schema table + `have_nz01` / `os.path.exists` guards) — which is why CI is green without the real-key CSV committed.
- **Added test** `tests/test_nz01_produced_cells.py` (ab2c248), mirroring `test_joint_four_cell.py`: the loader's modality-exclusive column mapping + zero-produced drop, and `prepare_joint_data`'s inclusion behind `include_nz01_cells` without double-counting the production-only marginal.

Outstanding items are correctly flagged as follow-ups, not blockers: depends on `research-data-analysis#8` + the real-key data commit; and the 675-vs-800 `n_trials` heterogeneity is pre-existing (shared with uk_03 = 418).

One incidental find while writing the test: the shared data-prep normality summary (`dse_research_utils` Anderson-Darling) raises `ZeroDivisionError` on an **all-NaN outcome column**. Not triggered here in practice (real data populates every outcome), but worth hardening in the descriptive helper.

PR #145, opened 2026-07-07
feat(scripts): add run_replication.sh full-replication runner

frankbuckley:
> [!NOTE]

## Summary

Adds `scripts/run_replication.sh`, a one-command, recoverable runner for full reporting-config replication runs (`--config rep`, ~15–25h across all registered models) on a remote VM. It productionises the ad-hoc orchestration used during a live replication run, folding in several failure modes hit the hard way.

## Why

Running `fit_model.py all --config rep --render --upload` directly has three sharp edges on a long remote run: an SSH disconnect kills the whole job; a single model raising aborts every remaining fit; and a broken blob upload credential fails the run *after* the expensive sampling. This script addresses all three.

## What it does

Phases (each independently toggleable): data prep + descriptives → fit every model (rendered) → read-only comparisons → sync figures + render reports → upload model output.

Key properties:
- **`--detach`** re-execs under `setsid` so an SSH disconnect can't kill the run. The script activates conda explicitly, since detached (non-login) shells don't — otherwise `python` resolves to the wrong interpreter and every step fails instantly.
- **Per-model isolation**: each model fits in its own subprocess, so one failure is logged and skipped rather than aborting the whole run (unlike the in-process `fit_model.py all` loop).
- **Idempotent / resumable**: a model whose `trace.nc` already exists is skipped (`--fresh` to override), so a rerun resumes cleanly after any interruption.
- **Upload decoupled** to a final phase and forced to use the az CLI credential via `AZURE_TOKEN_CREDENTIALS=dev` — `DefaultAzureCredential` otherwise picks the VM managed identity, which lacks the blob data role (`AuthorizationPermissionMismatch`).
- **Logging**: timestamped, tee'd `run.log` + per-step `status.tsv` + a `latest` symlink under `<output>/replication-logs/`.
- Model list is read from `MODEL_REGISTRY` (single source of truth; auto-adapts to new models).

Flags: `--config`, `--models`, `--output-dir`, `--log-dir`, `--env`, `--fresh`, `--include-kfold`, `--no-{descriptives,fit,compare,render,upload}`, `--detach`.

## Usage

```bash
scripts/run_replication.sh --detach            # full run, survives disconnect
tail -f output/replication-logs/latest/run.log # follow progress
```

## Testing

- `bash -n` syntax check passes.
- `--help`, registry enumeration, phase gating, logging/status/trap, and the `--detach` self-re-exec path (no orphan run dir) all validated with all-phases-off smoke runs.
- The equivalent inline pipeline is currently driving a live reporting-config replication run.

### DIFF
--- a/scripts/run_replication.sh
+++ b/scripts/run_replication.sh
@@ -1,0 +1,231 @@
+#!/usr/bin/env bash
+# Copyright (c) 2026 Down Syndrome Education International and contributors
+# SPDX-License-Identifier: AGPL-3.0-or-later
+#
+# run_replication.sh — one-command, recoverable full replication run.
+#
+# Fits every registered model at a chosen sampling config, runs the read-only
+# comparisons, syncs report figures, renders the reports, and uploads model
+# output to blob storage. Built for long reporting-config runs (~15-25h) on a
+# remote VM, so it is resilient to SSH disconnects and mid-run interruptions.
+#
+# Key properties
+#   * Detached: --detach re-execs under setsid so an SSH drop can't kill it.
+#   * Conda-safe: activates the env explicitly (setsid shells don't do this).
+#   * Per-model isolation: each model fits in its own subprocess, so one
+#     failure is logged and skipped rather than aborting the whole run.
+#   * Idempotent / resumable: a model whose trace.nc already exists is skipped
+#     (override with --fresh). Re-running after any interruption resumes.
+#   * Upload decoupled from fitting: a broken blob credential can never sink a
+#     multi-hour fit. Uploads run last, using the az CLI credential (see below).
+#   * Full logging: timestamped, tee'd terminal log + a per-step status TSV,
+#     with a "latest" symlink to the current run's log dir.
+#
+# Blob credential note: DefaultAzureCredential picks the VM managed identity
+# (which lacks the blob data role) before the az CLI login, giving
+# AuthorizationPermissionMismatch. We export AZURE_TOKEN_CREDENTIALS=dev so it
+# uses the interactive az login instead. Requires a valid `az login`.
+
+set -uo pipefail
+
+usage() {
+  cat <<'EOF'
+Usage: scripts/run_replication.sh [options]
+
+  --config <c>       Sampling config: dev | test | rep | rep-lite (default: rep)
+  --models "<...>"   Space-separated model keys to run (default: all registered)
+  --output-dir <d>   Output root (overrides $DSE_VOCAB_GROWTH_OUTPUT_DIR)
+  --log-dir <d>      Where to write run logs (default: <output>/replication-logs)
+  --env <name>       Conda env name (default: dse-vocab-growth)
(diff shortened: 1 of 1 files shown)


frankbuckley:
> [!NOTE]

## Validated by a full reporting-config run

`run_replication.sh` (this PR) was exercised end-to-end on a complete `--config rep` replication of all 15 registered models, plus comparisons, report render, and blob upload. Every design property in the script earned its place during the run: `setsid` detach survived two SSH disconnects; explicit conda activation avoided a silent `ModuleNotFoundError` in the detached shell; per-model isolation + idempotent `trace.nc`-based resume let the run recover cleanly after an interruption and a mid-run reconfiguration; and decoupled upload with `AZURE_TOKEN_CREDENTIALS=dev` (managed-identity → az-CLI credential) kept a broken blob credential from sinking the multi-hour fits.

## Findings worth a follow-up (not in this PR)

Two things surfaced that this runner does not yet address:

1. **TD random-effects models OOM at full `rep` on the 125 GB VM.** VG11 (`age-spoken-td-re`) grew to ~123 GB and was OOM-killed after ~8h with no output; per-observation posterior/predictive arrays scale with N×draws, and TD's N≈4,138 (3.5× the DS models) blows past memory at 36,000 draws. The DS RE models and single-outcome TD models fit fine; VG12 squeaked in just under the limit. Workaround used: fit the heavy TD models (VG11, VG13) at `rep-lite` (≈54 GB, still clears ESS>400) while keeping everything else at full `rep`. The runner currently takes a single `--config` for the whole fleet; a **per-model config map** (rep-lite for OOM-prone TD models) would make this reproducible.

2. **arviz 1.2 broke the comparison scripts.** `loo_compare.py` and `loso_compare.py` needed fixes for the 1.x API: `ELPDData.elpd_loo`→`.elpd` / `.p_loo`→`.p`, `az.compare` dropping its `ic=` argument, and `az.from_netcdf` now returning a `DataTree` whose `.groups` is a path-style property (`/log_likelihood`). Fixes are in the working tree; they can go in this PR or a separate one — say which you prefer.

## Run outcome

All 15 models fitted and uploaded. Convergence is clean throughout (R̂=1.0, min ESS≥450); divergences only on VG13 (5/16,000), VG02, VG12 (1 each). PSIS-LOO Pareto-k is clean everywhere except VG16 (cross-lag: 244/1903 k>0.7, LOO unreliable — expected). LOSO marginal favours VG09 > VG08 > VG07; the RE structure (VG07) beats no-RE (VG05) on LOO.

Refreshed artifacts (this run): [comparison report](https://dseresearch.blob.core.windows.net/public/projects/vocabulary-growth/output/019f42ee-4f12-77b5-8e91-1c79ee88affd/comparison-report/index.html) · [comparisons + LOO/LOSO tables](https://dseresearch.blob.core.windows.net/public/projects/vocabulary-growth/output/019f42ee-595e-75d8-b0df-aa03211d0138/comparisons/).

PR #170, opened 2026-07-18
refactor: harden model fitting and replication

### LINKED ISSUE #169: Improve model reproducibility and undergraduate readability
> [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

## Problem

The registered statistical models are coherent and well tested, but the surrounding workflow currently weakens reproducibility and makes the implementation harder to study than the statistical report. In particular, the replication script treats any `trace.nc` as a completed fit even when it came from a development run or was retained after a convergence failure; a new fit deletes the preceding successful output before its replacement passes; upload and report synchronisation do not independently verify publishability; dependency ranges are auditable through the fit manifest but are not locked for exact reconstruction; reporting query ages help determine the HSGP approximation domain; and the model-building code remains difficult for undergraduate readers to trace from equation to PyMC graph.

## Goals

- Make fitting, resuming, synchronising and uploading fail closed around an explicit definition of a complete, publishable fit.
- Preserve previous successful artefacts until a replacement has passed every required stage.
- Make published computational environments reconstructable as well as auditable.
- Separate statistical model settings from reporting-only settings.
- Make every registered model definition validate early and build in CI.
- Provide a worked undergraduate path from a simple count model to the joint hierarchical model.
- Record, rather than silently introduce, statistical extensions that require model comparison and reporting-quality refits.

### 1. Fit identity and lifecycle

- Add a machine-readable fit lifecycle with explicit `prepared`, `sampled`, `converged` and `complete` states.
- Define compatibility from model definition, sampling tier and parameters, prepared-data hash, and code revision/dirty state.
- Resume only a compatible complete fit; never accept file presence alone.
- Treat `CONVERGENCE_FAILED.txt` and incomplete lifecycle state as non-publishable.
- Record success and failure distinctly in replication-run status rather than writing an unconditional `DONE` marker.

### 2. Atomic, publication-safe artefacts

- Fit into a staging or immutable run directory.
- Promote a run to the canonical publishable location only after diagnostics, posterior prediction, summaries, plots and report generation succeed.
- Retain the previous publishable run if a replacement fails.
- Require upload and report-figure synchronisation to validate the manifest and lifecycle state.
- Add regression tests for development traces, convergence-failure traces, stale manifests, interrupted runs and successful promotion.

### 3. Reconstructable environment

- Add a documented release-lock workflow and platform lock artefacts, while keeping `environment.yml` as the maintainable canonical specification.
- Check that the lock and canonical environment specification agree in CI.
- Reconcile package constraints that currently differ between `environment.yml`, `pyproject.toml` and contributor guidance.

### 4. Model/readability separation

- Add an explicit HSGP model domain independent of plot and query ages, and validate that every evaluation/anchor age lies inside it.
- Add definition validation for ordered anchors, positive prior parameters, compatible feature flags, unique registry names and valid age windows.
- Replace misleading placeholder outcome storage in multi-outcome preparation with typed prepared-data structures where this can be done without changing graphs.
- Move refactor history out of primary statistical docstrings and correct stale model descriptions.
- Add a parametrised registry-wide model-build test.

### 5. Undergraduate learning path

- Add an executable Quarto walkthrough using synthetic data: Binomial count, Beta-Binomial heterogeneity, anchored logistic age trend, GP departure, study/child effects, and the nested `U` plus `S | U` model.
- Complete the equation-to-code glossary for `f`, `g`, `h`, `p_U`, `q`, `eta`, `ell` and `kappa`.
- Explain the distinction between population trajectories and predictions for a new child.

### 6. Statistical follow-ups requiring new fits

These should be implemented and evaluated separately from the workflow/readability refactor because they change the model and require sensitivity comparisons and reporting-quality refits:

- correlated study and child random effects across understood, production-ratio and signing trajectories;
- a free-sign or constant-dispersion alternative to the forced monotone `kappa` trend;
- continued cautionary treatment of the VG16 within-child lag, whose short-panel identification limitation is data-driven.

## Acceptance criteria

- A default reporting replication never skips a development-tier, failed, stale or incompatible trace.
- Failed replacement fits leave the previous publishable artefacts intact.
- Upload and figure synchronisation reject non-publishable runs.
- Published environment snapshots can be recreated for supported platforms.
- Plot/query changes inside a fixed HSGP domain do not change the model specification.
- Every registered model definition validates and builds in CI.
- The undergraduate walkthrough renders and maps its equations directly to the production variable names.
- Existing unit, lint and formatting checks pass, with new lifecycle/state-machine coverage added.

frankbuckley:
> [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

## Summary

This implements the reproducibility and undergraduate-readability plan in #169.

- Treats a fit as an explicit lifecycle rather than the existence of `trace.nc`: models fit and optionally render in an isolated staging directory, successful output is atomically promoted, prior complete output survives a failed replacement, and failed staging output is retained for diagnosis.
- Adds one compatibility validator for resume, report-figure synchronisation, and publication. It checks lifecycle state, registered definition, sampling tier and parameters, raw-data fingerprint, Git revision and cleanliness, convergence marker, trace, and rendered report as appropriate.
- Makes replication fail closed: development, stale, interrupted, dirty-source, or incompatible fits cannot be skipped or uploaded; a failed required step prevents downstream comparisons and publication and leaves a `FAILED` marker.
- Separates the HSGP approximation domain from reporting query ages, records an explicit domain in every registered model, validates all declarative specifications at import, and smoke-builds every registered PyMC graph.
- Adds exact `linux-64` and `osx-arm64` Conda locks, an exact pip lock including the immutable shared-library commit, a lock-refresh workflow, and richer per-fit package-origin provenance.
- Adds an executable undergraduate model walkthrough, completes the glossary and Bayesian-workflow sections, and corrects stale dependency/model-count guidance.

The statistical structure and estimands are intentionally unchanged: correlated random effects, alternative dispersion functions, and interpretation of the VG16 cross-lag require separate sensitivity fits and are not silently introduced as readability refactors.

## Verification

- `pytest`: 217 passed (15 existing numerical/library warnings)
- Registry-wide PyMC graph build: all 15 registered models
- `ruff check src/ scripts/ tests/`
- `npm run format:check`
- `npm run spellcheck`
- `bash -n scripts/run_replication.sh`
- `dse-check-env environment.yml`
- Executed and rendered `docs/report/tutorial-model-walkthrough.qmd` to HTML
- `git diff --check`

Closes #169


src/vocab_growth/fit_artifacts.py
```diff
@@ -0,0 +1,349 @@
...
+    recorded_sampling_name = sampling_payload.get("configuration_name")
+    if (
+        expected_sampling_config_name is not None
+        and recorded_sampling_name != expected_sampling_config_name
+    ):
+        errors.append(
+            "Sampling configuration mismatch: "
+            f"found {recorded_sampling_name!r}, expected {expected_sampling_config_name!r}."
+        )
+    if expected_sampling_parameters is not None and sampling_payload.get(
```
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Fable 5).

**[1 — blocker] The sampling-parameter check requires exact dict equality, but the compared dict includes the host-dependent `cores` field and can never match intentional `refit_hightune.py` overrides.**

`cores` is `min(chains, physical_cores - 1)` (from `dse_research_utils.statistics.models.sampling`), so a rep fit made on a VM with ≥7 physical cores records `cores=6`, and validation on a 6-core laptop expects `cores=5` → "Sampling parameters differ from the current configuration" for every model, forcing a full ~15–25 h refit. Every consumer inherits this: `check_fit.py`, `sync_report_figures.py`, `upload.py`, and both `run_replication.sh` phases. Separately, a runbook-sanctioned `refit_hightune.py` fit (e.g. `--tune 12000`, per `docs/runbooks/full-refit.md`) writes a manifest that can never validate against the unpatched `rep` config, so the heavier-tuned model of record can never be synced, published, or skipped on resume.

Suggested fix: exclude host-dependent fields (`cores` at minimum) from the comparison, and define an acceptance policy for intentional tuning overrides (e.g. compare only tier-defining fields, or an explicit override marker in the manifest).

src/vocab_growth/models/common.py
```diff
@@ -1599,35 +1585,113 @@ def run_fit_pipeline(
...
+                config_name=definition.config_name,
+                sampling_config_name=config,
+                stage=name,
+            )
+            with section(name, timings=timings):
+                fn(context)
+                if stage_index == 0:
+                    write_fit_manifest(context, definition)
+
+        if render:
```
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Fable 5).

**[2 — blocker] Rendering now runs inside the staged try-block before promotion, so a render failure demotes a fully-sampled fit to `failed/`.**

A `CalledProcessError` (or `FileNotFoundError` if quarto is missing) falls into the `except BaseException` path, which writes `fit_state` "failed" and `retain_failed_fit()` moves the staged dir — trace, manifest, plots — to `<output_root>/failed/<label>-<timestamp>`. `check_fit.py` only inspects the canonical `models/` dir and `validate_fit_output` rejects `state != "complete"`, so a ~10 h rep fit becomes unreachable by every automated path and the next replication run refits from scratch instead of just re-rendering. On `main`, fits were promoted first and rendered afterwards in `fit_model.py`, so a render failure was recoverable.

Suggested fix: promote the completed fit first and render afterwards, or catch render errors, still promote, and record a distinct render-failed status that `--render`/publish paths can retry.

scripts/fit_model.py
```diff
@@ -23,22 +21,21 @@
...
+def _fit_selected_models(selected, config: str, *, render: bool = False):
     """Fit every selected model while collecting convergence-gate failures."""
     contexts = []
     timings: dict[str, float] = {}
     failures: dict[str, str] = {}
     for name, module in selected:
         model_started = time.perf_counter()
         try:
-            contexts.append(module.fit(config))
+            contexts.append(module.fit(config, render=render))
```
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Fable 5).

**[3 — major] Only `ConvergenceGateError` is caught, but render errors now surface from `module.fit()` itself, so one model's render failure aborts the rest of an `all` batch.**

With `fit_model.py all --config rep --render`, a quarto/pandoc failure on model 3 propagates out of `_fit_selected_models`, models 4–15 are never fitted, and (per the staging behaviour in `common.py`) model 3's completed trace lands in `failed/`. If quarto is missing entirely, model 1 fails at render time and nothing else is ever fitted — on `main`, all fits completed before the first render was attempted.

Suggested fix: catch render failures per model alongside the convergence-gate failures and report them in `failures` at the end.

src/vocab_growth/models/common.py
```diff
@@ -487,7 +494,7 @@ def extract_model_samples(trace: xr.DataTree) -> ModelSamples:
     return model_samples
 
 
-def build_model(context: ModelFitContext):
+def build_model(context: ModelFitContext, definition=None):
```
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Fable 5).

**[4 — major] `definition=None` falls back to the observed age range as the HSGP domain, and `scripts/prior_predictive_audit.py` still calls `build_model(ctx)` with no definition (lines 69/79/84), so the audit crashes for some models and silently audits a different HSGP approximation for others.**

`python scripts/prior_predictive_audit.py vg04` now raises: VG04's comprehension observations end at 25 months but `ages_query` reaches 30, so `construct_age_grids` rejects the query ages as outside the (observed-range) domain. For definition-less audits that don't crash — e.g. VG14 via `ct.build_model(ctx)`, which is in the default audit set — the HSGP `L`/`M` are sized from the observed range rather than the declared `(8, 115)` domain; it only coincides today because the DS pool happens to span exactly 8–115 months. The same optional-with-fallback pattern is copied in `common_bivariate.py:364` and `common_trivariate.py:479`.

Suggested fix: make `definition` required in all three engines (validate_model_registry already mandates `gp_domain_months` for every registered model) and pass `d` at the three audit call sites, as the joint branch (`cj.build_model(ctx, d)`) already does.

scripts/run_replication.sh
```diff
@@ -195,16 +206,28 @@ log "phases: descriptives=$DO_DESCRIPTIVES fit=$DO_FIT compare=$DO_COMPARE rende
...
     run_step "fit_$m" python scripts/fit_model.py "$m" --config "$CONFIG" --render
   done
+  stop_if_failed
+fi
+
+# Verify all inputs before read-only comparisons or publication. This also
+# protects --no-fit runs from consuming stale or development-quality traces.
+if [ "$DO_COMPARE" = 1 ] || [ "$DO_RENDER" = 1 ] || [ "$DO_UPLOAD" = 1 ]; then
+  for m in $MODELS; do
+    run_step "validate_$m" python scripts/check_fit.py "$m" --config "$CONFIG" \
```
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Fable 5).

**[5 — major] The dirty-checkout error fires under `--purpose resume` (it is inside the `expected_git` block in `validate_fit_output`, not gated on `require_clean_fit`), and nothing checks cleanliness before the 15–25 h fit phase.**

A run started with one untracked scratch file completes every fit, then each `validate_<m>` step fails with "The current checkout is dirty, so an exact resume is unsafe" and `stop_if_failed` aborts before comparisons/render/upload, marking the run FAILED despite all fits having succeeded. The same condition makes `has_compatible_fit` exit non-zero (the error is only visible in `run.log`), so an interrupted-and-resumed run silently refits models that already completed.

Suggested fix: preflight git cleanliness at the top of the script (fail fast, before fitting), and gate the dirty/commit checks in `validate_fit_output` on purpose rather than on `expected_git`'s presence.

scripts/upload.py
```diff
@@ -68,6 +76,20 @@
...
             exit(1)
 
+        definition = MODEL_REGISTRY[model_id]
+        expected_sampling = sampling.get_sampling_configuration(args.config)
+        require_valid_fit(
+            output_dir,
+            expected_definition=definition,
+            expected_sampling_config_name=args.config,
+            expected_sampling_parameters=asdict(expected_sampling),
+            expected_git=git_metadata(local_env.ROOT_DIR),
```
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Fable 5).

**[6 — major] Passing `expected_git` makes publishing require the current HEAD to exactly equal the fit's commit, so any post-fit commit — even docs-only, or merging this PR itself — blocks upload of a valid rep fit.**

`validate_fit_output` hard-errors on commit mismatch whenever `expected_git` is provided, and the only recovery is a multi-hour refit. This is stricter than the PR's own documented policy: the report text describes the Git-revision match as a *resume* requirement, while upload requires reporting quality, a clean checkout, and a rendered report. The adjacent error message ("…an exact resume is unsafe") suggests this validator branch was designed for resume and reused wholesale in the publish path.

Suggested fix: for publish, drop the commit-match expectation (keep the clean-checkout requirement via `require_clean_fit`, which checks the manifest's recorded state).

scripts/sync_report_figures.py
```diff
@@ -81,15 +100,34 @@ def main() -> None:
...
             for name in sorted(os.listdir(models_dir)):
                 src = os.path.join(models_dir, name)
-                if os.path.isdir(src):
-                    n = _sync_dir(src, os.path.join(env.REPORT_FIGS_DIR, name))
-                    total += n
-                    print(f"  {name}: {n} files")
+                definition = definitions_by_label.get(name)
+                if not os.path.isdir(src) or definition is None:
+                    continue
+                require_valid_fit(
```
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Fable 5).

**[7 — major] `--allow-provisional` only relaxes `require_reporting_quality`/`require_clean_fit`; `expected_git` and `expected_source_data_hash` are still passed unconditionally, so the flag fails in exactly the local-dev situation it exists for.**

A developer with any uncommitted edit (or any commit made after fitting) who runs `sync_report_figures.py --config dev --allow-provisional` hits the dirty-checkout or commit-mismatch error from `validate_fit_output` and syncs nothing, despite the flag's stated purpose ("Allow complete dev/test fits in the local cache").

Suggested fix: skip the git (and possibly data-hash) expectations when `--allow-provisional` is set.

frankbuckley (commented):
> [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Fable 5).

## Review summary

This review ran 8 finder passes (line-by-line, removed-behaviour, cross-file tracing, reuse, simplification, efficiency, altitude, conventions) followed by an adversarial verification pass; every finding below was confirmed against the actual code on this branch, not just the diff. The provenance-validation layer in `fit_artifacts.py` is a sound design, but several checks are stricter than the workflows the repo documents, and in the documented fit-on-VM / sync-on-laptop setup they will reject virtually every legitimate fit. Ten confirmed findings are posted as inline comments, ranked by severity; three root causes drive most of them.

**1. The sampling-parameter equality check compares host-dependent and intentionally-overridden values.** `cores` comes from `min(chains, physical_cores - 1)`, so a fit made on one machine fails validation on any machine with a different core count; and `refit_hightune.py` fits (the documented model of record for ridge models) can never match the unpatched `rep` configuration. Every consumer — `check_fit.py`, `sync_report_figures.py`, `upload.py`, `run_replication.sh` resume/validate — inherits this.

**2. The `expected_git` block errors unconditionally on a dirty current checkout or any post-fit commit**, regardless of purpose or flags. This makes `--allow-provisional` unusable for its stated purpose, blocks `upload.py` from publishing a valid rep fit after any later commit (stricter than the PR's own documented policy), silently defeats `run_replication.sh` resume, and — with no cleanliness preflight before the 15–25 h fit phase — lets a dirty-checkout replication run fit everything and then fail every validate step.

**3. Quarto rendering moved inside `run_fit_pipeline`'s try-block, before promotion.** A render failure now marks a fully-sampled multi-hour fit `failed` and moves it to `failed/`, where no resume path can reach it; on `main`, fits were promoted first and rendered afterwards. Combined with `fit_model.py` catching only `ConvergenceGateError`, one render failure aborts an entire `all` batch.

### Also confirmed (below the inline-comment cut)

- `fit_model.py --config test --render --upload` passes argparse validation, then crashes after the fit inside `storage.upload_to_blob_storage`'s embedded `require_valid_fit(require_reporting_quality=True)`; the reporting-quality precondition should be checked up front like the `--upload requires --render` one. Relatedly, the `fit_model.py --upload` path and `scripts/upload.py` now enforce two different publish gates (the embedded check in `storage.py` has no definition/sampling/git/data-hash expectations), and `upload.py` validates the same directory twice.
- The expected-provenance kwarg block is hand-copied in `check_fit.py`, `sync_report_figures.py`, and `upload.py`, and they already disagree: sync omits `require_rendered_report`, so an unrendered fit syncs into the report cache but fails publish. A single `fit_artifacts` helper (purpose → kwargs) would prevent the drift. Same story for the duplicated `_json_default` and `git_metadata` copies living on opposite sides of the manifest-write/validate boundary — and `fit_manifest.json`, the one file every validator parses, is still written non-atomically while `_write_json_atomic` sits unused beside it.
- Efficiency: `upload.py all` recomputes `git_metadata` plus a full ~13 MB `source_data_hash` inside the per-model loop; `sync_report_figures.py --comparisons-only` computes both and never uses them; `run_replication.sh` spawns `check_fit.py` up to ~30 times per run (one interpreter, registry import, and data re-hash per model per phase) because it accepts only one model id.
- The new `render` flag is threaded as identical pass-through boilerplate through all 15 model wrappers plus 6 engine signatures; `**kwargs` forwarding (or registry-level dispatch) would collapse it, and a wrapper that forgets `render=render` fails silently on a render-only run.

### Conventions

Clean across the board: `CLAUDE.md`/`AGENTS.md`/`.github/copilot-instructions.md` are byte-identical, all new Python files carry the SPDX header, new docs carry AI attribution, prose is not hard-wrapped, and cspell/ruff/prettier all pass.

### Highest-leverage fix

Make `validate_fit_output`'s strictness purpose-aware: exclude `cores` (or all host/tuning-override fields) from the sampling comparison, gate the dirty/commit-match errors on the purpose or an explicit flag rather than on `expected_git`'s mere presence, and move rendering back outside the staged try-block (or catch render errors and still promote). That resolves findings 1, 2, 3, 5, 6, and 7 at the root rather than per-script.

scripts/run_replication.sh
```diff
@@ -215,20 +238,23 @@ if [ "$DO_COMPARE" = 1 ]; then
   for c in $CMP; do
     run_step "cmp_$c" python "scripts/$c.py"
   done
+  stop_if_failed
 fi
 
 # 4. Sync figures into the report cache, then render.
 if [ "$DO_RENDER" = 1 ]; then
-  run_step "sync_figures"      python scripts/sync_report_figures.py
+  run_step "sync_figures"      python scripts/sync_report_figures.py --config "$CONFIG"
```
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Fable 5).

**[8 — major] The usage text advertises `--config dev | test | rep | rep-lite`, but the sync step never passes `--allow-provisional`, and the new sync unconditionally requires reporting quality — so every dev/test replication run now fails at `sync_figures`.**

Fits and validate steps pass for `--config test` (resume purpose doesn't require reporting quality), then sync fails with "Sampling configuration 'test' is not reporting-quality", `stop_if_failed` halts before upload, and the run leaves a FAILED marker. On `main`, the sync script performed no validation, so a test-tier end-to-end rehearsal previously worked. (`rep-lite` is unaffected — it is in `REPORTING_CONFIGS`.)

Suggested fix: pass `--allow-provisional` when `$CONFIG` is non-reporting, or reject non-reporting configs up front in the usage/argument check.

scripts/sync_report_figures.py
```diff
@@ -81,15 +100,34 @@ def main() -> None:
...
+        f"{definition.model_id}-{definition.config_name}": definition
+        for definition in MODEL_REGISTRY.values()
+    }
+    expected_sampling = sampling.get_sampling_configuration(args.config)
+    current_git = git_metadata(env.ROOT_DIR)
+    current_source_hash = source_data_hash(env.DATA_DIR)
 
     if not args.comparisons_only:
         if os.path.isdir(models_dir):
             for name in sorted(os.listdir(models_dir)):
```
frankbuckley, line 140: > [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Fable 5).

**[9 — moderate] The loop validates-then-copies per directory with no error handling, so the first stale model dir aborts the sync mid-loop, leaving `docs/report/figures/` mixing new and old figures, and the comparisons phase never runs; non-registry directories are silently skipped with no log line.**

With fresh rep fits for most models plus one leftover dev fit, alphabetically earlier models are copied, then `require_valid_fit` raises and `main()` exits — a mixed cache is worse than either all-old or all-new. The silent `continue` also drops `fit_sensitivity.py`'s `<label>-<suffix>` output dirs, which the old sync copied (nothing in the report consumes them today, but the skip deserves a log line).

Suggested fix: validate all model dirs first and copy only if all pass (or collect errors and report them at the end, as `check_fit.py` already does), and log skipped directories.

src/vocab_growth/models/common.py
```diff
@@ -1198,30 +1208,13 @@ def run_standard_plots(context: ModelFitContext, *, outcome_label: str = "Word c
 
 def report(context: ModelFitContext):
```
frankbuckley, line 1211: > [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Fable 5).

**[10 — moderate] `report()` no longer clears `docs/report/figures/<model>/` (on `main` it `rmtree`'d before copying), and the now-sole refresh path `_sync_dir` in `sync_report_figures.py` only copies without clearing — so figures/CSVs renamed or removed between fits linger in the report cache forever.**

Rename a plot (`q_ratio.svg` → `q_ratio_by_age.svg`), refit, and sync: both files exist in the cache, and any `.qmd` still referencing the old name silently renders the stale figure from the previous fit instead of failing.

Suggested fix: have `_sync_dir` replace the destination directory (clear-then-copy) rather than overlay it.

src/vocab_growth/fit_artifacts.py
```diff
@@ -0,0 +1,349 @@
...
+
+    path = os.path.join(output_dir, FIT_STATE_FILENAME)
+    now = datetime.now(UTC).isoformat()
+    started_at = now
+    if os.path.isfile(path):
+        try:
+            started_at = read_json(path).get("started_at_utc", now)
+        except FitValidationError:
+            pass
+
```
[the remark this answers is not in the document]
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

Addressed in `5d4ab39`. `write_fit_state()` now reads and retains the previous stage whenever a terminal write omits `stage`. The regression test writes `running` at `Posterior sampling`, then `failed`, and verifies that the failure metadata still names that stage.

src/vocab_growth/models/definitions.py
```diff
@@ -1117,3 +1158,117 @@ def model_type(self) -> ModelType:
...
+                raise ValueError(f"{field_path} must be positive; got {field_value!r}.")
+
+
+def validate_model_definition(definition) -> None:
+    """Fail early when a declarative model specification is internally invalid."""
+    prefix = getattr(definition, "model_id", type(definition).__name__)
+    if not re.fullmatch(r"VG\d{2}", definition.model_id):
+        raise ValueError(f"{prefix}.model_id must have the form VG01.")
+    if not definition.config_name or any(char.isspace() for char in definition.config_name):
+        raise ValueError(f"{prefix}.config_name must be a non-empty path-safe label.")
```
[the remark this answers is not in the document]
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

Addressed in `5d4ab39`. `config_name` now has an allowlist-style validation (`A–Z`, `a–z`, digits, `_`, and `-`, with alphanumeric endpoints), which rejects `/`, `\\`, `..`, and other path components. Parameterised tests cover both separator forms and parent-directory traversal.

src/vocab_growth/fit_artifacts.py
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

Addressed in `5d4ab39`. Compatibility now ignores host-dependent `cores`; requires `draws`, `tune`, `chains`, and `target_accept` to meet or exceed the named tier; and continues to compare other settings such as the random seed exactly. Tests verify that a 12,000-tune/8,000-draw/0.99-target `rep` fit with different cores is accepted, while a weaker fit is rejected.

src/vocab_growth/models/common.py
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

Addressed in `5d4ab39`. Rendering has been removed from the staged statistical pipeline. A successful fit is marked complete and atomically promoted first; Quarto then renders the canonical output. `--render-only` validates and retries an existing fit without sampling, so a Quarto failure cannot demote or hide a completed trace.

scripts/fit_model.py
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

Addressed in `5d4ab39`. Fitting and rendering are now separate batch phases: every model is fitted before rendering starts, and `_render_contexts()` catches and records each model’s render error while continuing with the remainder. A regression test confirms first/bad/last all render attempts occur and the single failure is reported.

src/vocab_growth/models/common.py
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

Addressed in `5d4ab39`. `definition` is now required by the univariate, bivariate, and trivariate builders, and every prior-audit call passes the registered definition. I also executed `prior_predictive_audit.py vg04` against an isolated output root; it completed successfully using the declared GP domain.

scripts/run_replication.sh
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

Addressed in `5d4ab39`. The replication script now checks Git cleanliness immediately after resolving the repository and before creating logs or fitted output. Validation policy is explicit by purpose: resume requires a clean current checkout plus matching commit/data, while other consumers do not inherit those checks merely because Git metadata was supplied.

scripts/upload.py
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

Addressed in `5d4ab39`. Publish validation no longer receives or compares the current Git commit. It still requires a reporting-quality, convergence-passing, rendered fit whose manifest records a clean source revision and whose registered definition and raw-data fingerprint remain compatible. A regression test verifies that a later current commit does not invalidate publication.

scripts/sync_report_figures.py
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

Addressed in `5d4ab39`. `provisional-sync` is now a distinct validation policy: it retains complete lifecycle, registered-definition, requested-tier, and minimum-sampling checks, but deliberately omits current Git, raw-data, reporting-quality, and clean-fit requirements. This makes the flag usable for its stated local-development purpose.

scripts/run_replication.sh
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

Addressed in `5d4ab39`. `run_replication.sh` classifies the sampling tier up front, passes `--allow-provisional` for dev/test figure syncs, and disables publication for those non-reporting runs. `rep` and `rep-lite` keep the strict sync and upload policy.

scripts/sync_report_figures.py
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

Addressed in `5d4ab39`. The sync now validates every registered model directory and reports all failures before changing any cache directory; unregistered and non-directory entries are logged explicitly. A regression test recreates the old valid-then-stale ordering and verifies that `_sync_dir` is never called when any validation fails.

src/vocab_growth/models/common.py
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

Addressed in `5d4ab39`. `_sync_dir` now builds an allowlisted snapshot in a sibling staging directory and atomically replaces the destination, restoring the prior directory if promotion fails. Tests verify that removed/renamed artefacts do not linger and that a copy failure preserves the previous cache.

frankbuckley:
> [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

Addressed the review summary in `5d4ab39`, including the findings below the inline-comment cut:

- `fit_model.py --upload` now rejects non-reporting configurations before fitting, and both upload entry points use the same purpose-derived publication policy.
- Batch uploads validate the complete set once before publishing any model; a validated-output token prevents the storage layer from applying a second, weaker or divergent gate.
- Resume, render, strict sync, provisional sync, and publish requirements now come from one `fit_validation_kwargs()` policy helper.
- Manifest writing now reuses the shared Git/JSON implementation and writes `fit_manifest.json` atomically.
- Upload hashes raw data once per batch; comparisons-only sync computes no model provenance; post-fit replication validation checks all selected models in one process.
- The `render` pass-through was removed from all 15 wrappers and six engines; rendering is now an explicit post-promotion operation with `--render-only` retry support.
- Strict figure sync requires rendered model output, validates the full input set before copying, and atomically replaces each cache directory.

Verification on the pushed commit: `pytest -q` (231 passed; 15 existing numerical/library warnings), registry-wide graph builds, successful isolated `prior_predictive_audit.py vg04`, Ruff, Prettier, CSpell, Bash syntax, synchronised agent-instruction copies, and `git diff --check`.

Issue #163, opened 2026-07-16
Review and resolve statistical-model validity risks before publication


frankbuckley:
> [!NOTE]
> Drafted by an LLM-based AI tool (Codex/GPT-5).

## Summary

A code-, data-, and documentation-level review of all 15 active registered models (VG01–VG05 and VG07–VG16) finds that the core trajectory framework is statistically coherent for exploratory population-average age associations, but the current methodology is not yet publication-sound for the headline signing, child-level understood-to-spoken, TD-comparison, or directional cross-lag claims.

The strongest parts are the bounded overdispersed Beta-Binomial formulation, smooth HSGP trajectories, interpretable anchor priors, non-centred study/child effects in the later DS models, exclusion of the Wordbank WS comprehension proxy, and generally strong recorded NUTS diagnostics. The full automated test suite and Ruff checks passed during the review. These establish implementation quality and successful sampling of the specified posterior; they do not establish that the model assumptions or interval coverage are correct.

### P0 — signing outcomes are not harmonised

UK01 records `signed` as **signed-only** words, excluding words that are also spoken. UK02 and NZ01 record **total signed** vocabulary, including words also spoken. The discrepancy is documented in [`data/vocab_data_uk_01.md`](https://github.com/dseinternational/vocabulary-growth/blob/main/data/vocab_data_uk_01.md#L124-L153), but [the combined-data view](https://github.com/dseinternational/vocabulary-growth/blob/main/src/vocab_growth/data_utils.py#L142-L152) passes the UK01 field through as the same outcome used elsewhere.

UK01 contributes 219 signed observations and dominates the late-age signing data; 114 UK01 signed observations are above 60 months. A study intercept cannot correct an age- and speech-overlap-dependent change in what the outcome means. This threatens VG14/VG15 estimates of the signing ratio, signing peak/decline, sign–speech overlap, and total expressive vocabulary. UK06 comparability is also uncertain.

**Required review:** harmonise UK01 from source item-level data if possible; otherwise compare pre-specified source-restricted fits, including exclusion of UK01 and UK06, before interpreting any signing result.

### P1 — the understood/spoken likelihood does not support the reported child-level conditional claim

[...]

Issue #65, opened 2026-06-17
Study random intercepts in common_bivariate_re use a centred parameterisation (non-centred elsewhere)


frankbuckley:
Flagged during the review in #64 (left unchanged there because it alters validated fits).

## Observation
`src/vocab_growth/models/common_bivariate_re.py` (`build_model_re`, ~lines 461–465) parameterises the **study** random intercepts in *centred* form:

```python
tau_u = pm.HalfNormal("tau_u", sigma=definition.tau_u_sigma)
delta_u = pm.Normal("delta_u", mu=0, sigma=tau_u, dims="study_id")
tau_q = pm.HalfNormal("tau_q", sigma=definition.tau_q_sigma)
delta_q = pm.Normal("delta_q", mu=0, sigma=tau_q, dims="study_id")
```

Everything else in the codebase uses the **non-centred** form `delta = tau * z`, `z ~ Normal(0, 1)`:
- the **subject** REs in the *same file* (`delta_subj_u = tau_subj_u * delta_subj_u_raw`),
- `common_univariate_re.py` (`delta = tau * delta_raw`),
- `common_joint_modality.py` (`delta_u = tau_u * z_u`, etc.).

## Why it matters
With few groups (the DS studies number well under ~10), the centred parameterisation produces the classic hierarchical "funnel" geometry, which HMC samples poorly. This may be connected to the `r_hat` / `ess_tail` issues on the q-trajectory hyperparameters that motivated VG09 → VG10 (tighter priors + GP anchoring). Switching to non-centred is mathematically the **same model** (identical posterior) but with friendlier geometry.

## Why it was not done in #64
It changes the trace and introduces new variable names (`delta_u_raw`, etc.), so it requires **re-fitting and re-validating VG07–VG13**. That belongs in its own PR with a before/after comparison rather than bundled with low-risk fixes.

## Suggested acceptance criteria
- [ ] Switch study REs in `build_model_re` to non-centred (consistent with the subject REs in the same file).
- [ ] Re-fit VG07–VG13 at `dev` then `rep`.
- [ ] Compare `r_hat`, `ess_bulk`, `ess_tail`, divergences, and population trajectories before vs after; confirm the population-level estimands are unchanged within MC error and diagnostics are no worse (ideally better).

PR #176, opened 2026-07-22
fix(models): resolve hierarchical-model convergence ridges (sum-to-zero REs + orthogonal-GP anchor)

frankbuckley:
> [!NOTE]

## Summary

Model-conditioning fixes that resolve genuine identifiability ridges in the hierarchical models, found during the full reporting-config (`rep`) refit of `VG01`–`VG16`. The ridges were structural, not sampling effort — heavier tuning did not help (`vg12` at hightune got *worse*: R-hat 1.013 → 1.019) — so each fix reconditions the model rather than throwing draws at it.

**Updated after review** (commits `23f5ba1`, `71e46b4`): the initial versions of the GP and RE fixes were correct in intent but flawed in implementation; the four P1 findings are now addressed and the guards run in CI. See the per-thread replies for detail.

## Changes

1. **Orthogonal-GP anchor** (`gp_utils.py`). The anchored HSGP deviation is orthogonalised against its mean's basis so it carries only nonlinear curvature and cannot alias with `slope` / the anchors. Crucially: the projection coefficients are fitted on the **observed rows only** (a fixed model design) and applied to every row, so the plot/query reporting grid cannot leak into the observed-row likelihood; and the residual is then **pinned to zero at the reference-age row**, preserving the `anchor_g*_at_ref` point-anchor contract. The nuisance basis is mean-specific: `[1, z]` for the logit-linear trend, `[1]` for the free-intercept mean (a linear GP direction there is genuine signal), and the three tent hats for the peak mean. Affects anchored outcomes only (`vg10`–`vg13`, `vg15`); `anchor_idx is None` is unchanged.

2. **Sum-to-zero study random intercepts** (`common_univariate_re`, `common_bivariate_re`, `common_joint_modality`). Constraining the unit offsets to sum to zero removes the study-RE-mean vs intercept ridge. This is an **intentional identifiability constraint, not a prior-preserving reparameterisation**: it removes the group-mean degree of freedom (that is the ridge) and imposes a −1/K correlation. `ZeroSumNormal(1)` would also shrink each marginal to `tau²·(K−1)/K`; we rescale sigma by `sqrt(K/(K−1))` so each study effect's **marginal prior variance stays `tau²`** (its value before the change), leaving only the mean DOF removed. Keeps #65's non-centred `tau · raw` scaling and every public name.

3. **Sign study RE over the informed subset** (`common_joint_modality`). `delta_sign` is zero-summed over only the **sign-informed studies** (union of `idx_sign` / `idx_cells` / `idx_prod`), scattered into full `study_id` with 0 for uninformed studies — so the constraint removes the sign-intercept ridge instead of letting nuisance studies counterbalance a common shift. U and q stay global (audited: informed by every retained study via their direct and nested terms).

## Validation

- Full test suite green; new **data-free** synthetic unit tests (`test_gp_anchor_orthogonalisation.py`) assert anchor-zero, observed-row orthogonality, reporting-grid invariance, the intercept/tent basis behaviour, and the RE-rescale marginal variance — and run in CI (the earlier guards only ran when the prepared DuckDB was present, i.e. never in CI).
- The joint sign fix is verified end-to-end on VG15: of 12 studies, the 5 sign-informed are zero-summed and the other 7 are forced to exactly 0.
- **Reporting-quality refits are being regenerated** on the corrected code (the GP-anchor and sign changes alter the likelihood; the RE rescale shifts the prior scale). The previously-quoted `vg10`/`vg15` before/after metrics were from the superseded whole-grid form and will be replaced with numbers from the corrected fits. The full run — config choices, incidents, the review, and the corrected fixes — is documented in `notes/202607191614-full-refit-rep-hightune-run.md`.

## Status / scope

No model definitions or prior semantics change beyond the documented RE constraint. Affected study-RE and anchored models (`vg07`–`vg13`, `vg15`, `vg16`) are being re-fit before publication; the no-RE / no-anchor baselines (`vg01`–`vg05`, `vg14`) are unaffected. Relates to the #163 statistical-validity work and the #164 hierarchical-TD change.

https://claude.ai/code/session_019anXhhvM8P55HkdKWMBpLZ

### DIFF
--- a/notes/202607191614-full-refit-rep-hightune-run.md
+++ b/notes/202607191614-full-refit-rep-hightune-run.md
@@ -185,3 +185,12 @@ All 15 model reports uploaded via `scripts/upload.py all --config rep` (traces e
 | VG16  | https://dseresearch.blob.core.windows.net/public/projects/vocabulary-growth/output/019f93b2-2106-75b2-8ad6-5476b3bdac6a/VG16-age-understood-spoken-ds-re-subj-uq-crosslag/index.html |
 
 Note: `upload.py` publishes the per-model reports; the consolidated report book (`docs/report`) and comparison book (`docs/comparison`) are rendered locally but not part of this per-model upload path.
+
+## Phase C addendum — consolidated books uploaded (2026-07-24)
+
+The consolidated report book (`output/report`, HTML + PDF + DOCX) and the comparison book (`docs/comparison`, rendered HTML; Quarto source excluded) were uploaded via `upload_directory_to_blob_storage` (shared run id `019f93ff-4065-706e-a592-d9e6839411bc`):
+
+| artefact                 | index URL                                                                                                                                     |
+| ------------------------ | --------------------------------------------------------------------------------------------------------------------------------------------- |
+| Consolidated report book | https://dseresearch.blob.core.windows.net/public/projects/vocabulary-growth/output/019f93ff-4065-706e-a592-d9e6839411bc/report/index.html     |
+| Comparison analysis      | https://dseresearch.blob.core.windows.net/public/projects/vocabulary-growth/output/019f93ff-4065-706e-a592-d9e6839411bc/comparison/index.html |
--- a/notes/202607191614-full-refit-rep-hightune-run.md
+++ b/notes/202607191614-full-refit-rep-hightune-run.md
@@ -148,7 +148,7 @@ Implemented and unit-tested (full suite green; ruff clean). **No model definitio
 | ----- | --------- | --------- | ------- | --- | -------- | ----------- |
 | vg07  | PASS      | 1.0005    | 8434    | 0   | 0.78     | —           |
 | vg08  | PASS      | 1.0020    | 1899    | 0   | 0.54     | —           |
-| vg09  | PASS      | 1.0033    | 1684    | 0   | 0.50     | —           |
+| vg09  | PASS      | 1.0033    | 1684    | 0   | 0.49     | —           |
 | vg10  | PASS      | 1.0016    | 5020    | 0   | 0.50     | —           |
 | vg15  | PASS      | 1.0012    | 4451    | 0   | 0.59     | —           |
 | vg16  | PASS      | 1.0062    | 1630    | 0   | 0.50     | —           |
@@ -156,7 +156,7 @@ Implemented and unit-tested (full suite green; ruff clean). **No model definitio
 | vg13  | PASS      | 1.0041    | 1215    | 0   | 0.28     | BFMI ~0.28  |
 | vg11  | PASS      | 1.0058    | 621     | 2   | 0.40     | 2 div       |
 
-The DS six are pristine (0 div, BFMI ≥ 0.5). The TD trio's BFMI ≈ 0.28 is intrinsic to the age-varying dispersion posterior over the narrow young-TD window — the same soft caveat accepted last run, now reached at plain `rep`. vg10 (DS anchored) is well conditioned (min ESS 5020), confirming the corrected obs-only anchor matches the flawed whole-grid form it replaced. These numbers supersede the whole-grid VG10/VG15 metrics quoted earlier.
+The DS six are pristine (0 div, BFMI 0.49–0.78; vg09 0.495 and vg16 0.496 sit just under 0.5). The TD trio's BFMI ≈ 0.28 is intrinsic to the age-varying dispersion posterior over the narrow young-TD window — the same soft caveat accepted last run, now reached at plain `rep`. vg10 (DS anchored) is well conditioned (min ESS 5020), confirming the corrected obs-only anchor matches the flawed whole-grid form it replaced. These numbers supersede the whole-grid VG10/VG15 metrics quoted earlier.
 
 **Provenance gap found and fixed.** The first corrected re-fit recorded `dirty=True` on every manifest: `git_metadata` computes dirty from `git status --porcelain --untracked-files=normal`, and 68 untracked `docs/comparison/*.png|svg` figure copies (left over from an earlier render) were in the tree at launch — `.gitignore` covered `*.csv`/`index.html` but not the plots. Fixed in `d2f2b96` (ignore the plot copies + report LaTeX byproducts), cleaned the tree, and re-fit on a porcelain-clean tree so all manifests are `dirty=False`. Lesson: the "never fit on an unclean tree" rule includes untracked build artifacts, not just modified tracked files.
 
--- a/notes/202607191614-full-refit-rep-hightune-run.md
+++ b/notes/202607191614-full-refit-rep-hightune-run.md
@@ -112,3 +112,30 @@ Restart of the deliberately-paused 2026-07-17 run (`202607170935-full-refit-vm-r
 **Provenance note:** published set will span commits 3e6f61d (baselines, code-identical) + b563586 (reparam) + the branch HEAD after this note. All docs/reparam commits are code-equivalent for each model's engine.
 
 **Infra lessons captured:** (1) never write into the working tree during a reporting run (dirties in-flight manifests); (2) launch long runs under `systemd-run` (system scope), not bare `setsid`/`nohup` (logind `KillUserProcesses` kills the latter on logout); (3) the hierarchical TD hightune fits are memory-heavy — draws 12000 OOMs on 251 GB; keep draws ≤ 8000 and never run two heavy fits concurrently.
+
+## GP orthogonalisation added, then TD trio landed at hightune (2026-07-21 late → 2026-07-22)
+
+- **GP orthogonalisation (commit `5397561`).** To remove the trend-vs-GP redundancy structurally (rather than lean on hightune), the anchored HSGP was changed from a single-point anchor (`g_unit - g_unit[idx]`) to a **whole-grid** orthogonalisation: project the constant AND linear-in-age components out of `g_unit` over the full `X_all_z` grid. This improved the already-passing DS anchored models — **vg10** R-hat 1.0027 → 1.0013, min ESS 2135 → 5111; **vg15** R-hat 1.0035 → 1.0011, min ESS 2050 → 6428 (0 divergences both). (These before/after numbers are from the `5397561` whole-grid form; see the correction below — they are superseded and will be re-derived.)
+- **TD trio** refit at hightune on the reparam + GP-orthogonalisation code: all three cleared the hard gate. Final soft caveats: **vg12** BFMI 0.29, **vg13** BFMI 0.28 (both pass R-hat/ESS); vg11 clean. All 15 models then passed the hard gate (11 fully clean, 4 soft-caveat: vg01 3 div, vg02 1 div, vg12/vg13 low BFMI).
+- Phase B (comparisons + render) completed; PR #176 opened.
+
+## Statistical review of #176 (PR reviewer via Codex/GPT-5, 2026-07-23) — four P1 + two P2
+
+The review found the two core conditioning commits were **correct in intent but flawed in implementation**. All six points verified against the code and accepted:
+
+- **P1-A — the whole-grid orthogonalisation leaked the reporting grid into inference.** `X_all_z` stacks observations with plot points, query ages and the anchor row; computing the projection statistics (`g.mean()`, the `z`-slope) over that whole stack makes the _observed-row_ latent — hence the likelihood — depend on `n_plot` / `ages_query`. A reporting-only choice must never move the posterior.
+- **P1-B — the whole-grid form broke the `anchor_g*_at_ref` point-anchor contract** (`g` no longer passed through zero at the reference age, contradicting the docstrings, the `definitions` field and the build output), and for the `tent` mean it projected against only `[1, z]`, a smaller basis than the three-anchor tent spans.
+- **P1-C — the sum-to-zero RE change was mis-described as prior-preserving.** `ZeroSumNormal(1)` has covariance `I − J/K` (marginal `1 − 1/K`, a 25 % shrink at K=4, plus a −1/K correlation); `tau` unchanged does not restore it.
+- **P1-D — the sign study zero-sum spanned all studies, not the sign-informed subset.** Uninformed studies' `z_sign` could counterbalance a common shift among the informed ones, so the constraint need not remove the sign-intercept ridge.
+- **P2-A** — the new guard tests skip in CI (`pytest` runs before `prepare_data.py`). **P2-B** — this note stopped before the GP-orthogonalisation commit and final validation.
+
+## Corrected fixes on the branch (2026-07-23)
+
+Implemented and unit-tested (full suite green; ruff clean). **No model definitions or priors semantics beyond the intentional, now-documented RE constraint.**
+
+- **GP anchor (P1-A + P1-B), `gp_utils.py`.** Projection coefficients are now fitted on the **observed rows only** (`n_obs`, a fixed model design) and applied to every row, so plot/query grids cannot leak into inference; the residual is then **pinned to zero at the anchor row**, restoring the point-anchor contract. The nuisance basis is **mean-specific**: `[1, z]` for the logit-linear trend, `[1]` for the free-intercept mean (a linear GP direction there is genuine signal), and the **three tent hats** for the peak mean. New data-free unit tests (`test_gp_anchor_orthogonalisation.py`) assert anchor-zero, observed-row orthogonality, reporting-grid invariance, and the intercept/tent basis behaviour; they run in CI.
+- **Sum-to-zero rescale (P1-C).** `ZeroSumNormal` sigma is rescaled by `sqrt(K/(K−1))` so each study effect's marginal prior variance stays `tau²`; only the group-mean DOF (the ridge) and a −1/K correlation remain. Relabelled in code/PR as an intentional identifiability constraint, not prior-preserving. Applied in `common_univariate_re`, `common_bivariate_re`, `common_joint_modality`.
+- **Sign zero-sum over informed studies (P1-D), `common_joint_modality.py`.** `delta_sign` is now `ZeroSumNormal` over the **sign-informed studies only** (union of `idx_sign` / `idx_cells` / `idx_prod`), scattered into full `study_id` with **0 for uninformed studies**. Verified on VG15: of 12 studies, 5 are active and zero-sum, 7 are forced to exactly 0. U and q remain global (audited: informed by every retained study via their direct + nested terms).
+- **P2-A** — added the synthetic CI tests above; updated the integration test's assertions to the new (observed-row, point-anchored) contract.
+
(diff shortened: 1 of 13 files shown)


src/vocab_growth/models/gp_utils.py
```diff
@@ -258,7 +258,18 @@ def _gp_from_mean(
...
-        g_unit = g_unit - g_unit[anchor_idx]
+        # Orthogonalise the GP deviation against the linear trend: remove its
+        # constant AND linear-in-age components over the grid, so `g` carries only
+        # nonlinear curvature. The old single-point anchor (``g_unit -
+        # g_unit[anchor_idx]``) removed only the level trade-off with the
+        # intercept; the GP could still tilt linearly and alias with ``slope`` — an
+        # R-hat ridge on ``p_slope_*``/``slope``/``g_unit_hsgp_coeffs`` that heavier
+        # tuning does not fix (it worsened for vg12 at hightune). Zeroing the mean
+        # and linear slope of ``g_unit`` decouples it from intercept + slope.
+        z = X_all_z_data[:, 0]
```
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex / GPT-5).

**[P1] Decouple inference from the reporting grid.**

`X_all_z_data` contains the observations together with `n_plot` plot points, query ages, and the anchor row. Estimating the projection coefficients from this combined array means that changing `n_plot`, `ages_query`, or the anchor row changes the projected GP at the observed rows, and therefore changes the likelihood and posterior. Please calculate the projection on a fixed model design—such as the training ages or fixed domain quadrature—and then apply those coefficients to every evaluation row. An invariance test should verify that changing only plot/query grids leaves the observed latent trajectory unchanged.

src/vocab_growth/models/gp_utils.py
```diff
@@ -258,7 +258,18 @@ def _gp_from_mean(
...
+        zc = z - z.mean()
+        g_unit = g_unit - g_unit.mean()
```
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex / GPT-5).

**[P1] This no longer implements the configured reference-age anchor.**

The block checks whether `anchor_idx` is non-null but never indexes `g_unit`, so `g_unit[anchor_idx]` is generally not zero. That contradicts the `anchor_g*_at_ref` contract and the build output, which state that every draw passes through zero at 13, 19, or 54 months. It also changes how the anchor-probability priors map to the trajectory. Please either preserve the point anchor while removing the linear component, or explicitly replace/rename the contract and update the model definitions and prior interpretation. For `tent_and_gp`, projecting only against `[1, z]` is additionally incomplete because the three-anchor piecewise-linear mean has a larger basis.

src/vocab_growth/models/common_univariate_re.py
```diff
@@ -371,11 +371,15 @@ def build_univariate_re_model(
...
+        # Study-level random intercepts (non-centred, sum-to-zero)
         # ============================================================
 
+        # Sum-to-zero on the unit offsets (delta_raw) removes the intercept vs
+        # study-RE-mean ridge: with few studies an unconstrained mean trades off
+        # against the global intercept/slope (R-hat failure at rep-hightune). The
+        # tau * raw scaling keeps the funnel-avoiding non-centring of issue #65.
         tau = pm.HalfNormal("tau", sigma=definition.tau_study_sigma)
-        delta_raw = pm.Normal("delta_raw", mu=0.0, sigma=1.0, dims="study_id")
+        delta_raw = pm.ZeroSumNormal("delta_raw", sigma=1.0, dims="study_id")
```
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex / GPT-5).

**[P1] Treat this as a prior/model change, not a prior-preserving reparameterisation.**

The original independent `Normal(0, 1)` offsets have covariance `I`; `ZeroSumNormal(1)` has covariance `I - J/K`, so each marginal variance becomes `1 - 1/K` and the offsets are correlated and constrained. For four studies this reduces the conditional marginal variance by 25%; multiplying by the unchanged `tau` does not restore the original prior. PyMC documents this covariance explicitly: https://www.pymc.io/projects/docs/en/stable/api/distributions/generated/pymc.ZeroSumNormal.html. If the constraint is intentional, please update the PR/model documentation, provenance, and prior-predictive sensitivity claims. If exact equivalence is required, the removed group-mean degree of freedom and the corresponding intercept/prior transformation must be retained.

src/vocab_growth/models/common_joint_modality.py
```diff
@@ -878,18 +878,20 @@ def build_model(context: JointContext, definition: JointModelDefinition):
...
-            "delta_u", tau_u * pm.Normal("z_u", 0.0, 1.0, dims="study_id"), dims="study_id"
+            "delta_u", tau_u * pm.ZeroSumNormal("z_u", sigma=1.0, dims="study_id"), dims="study_id"
         )
         delta_q = pm.Deterministic(
-            "delta_q", tau_q * pm.Normal("z_q", 0.0, 1.0, dims="study_id"), dims="study_id"
+            "delta_q", tau_q * pm.ZeroSumNormal("z_q", sigma=1.0, dims="study_id"), dims="study_id"
         )
         delta_sign = pm.Deterministic(
-            "delta_sign", tau_sign * pm.Normal("z_sign", 0.0, 1.0, dims="study_id"), dims="study_id"
+            "delta_sign", tau_sign * pm.ZeroSumNormal("z_sign", sigma=1.0, dims="study_id"), dims="study_id"
```
frankbuckley: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex / GPT-5).

**[P1] Constrain only studies that inform this outcome.**

`study_id` is built from every retained study, whereas the signing likelihood and composition terms use only a subset. Studies with no signing/cell/produced contribution therefore receive uninformed `z_sign` values that can counterbalance a common shift among the informed studies. The global zero-sum constraint consequently does not ensure that the sign-informed effects themselves sum to zero, so it need not remove the intended sign-intercept ridge and it introduces coupling through nuisance effects. Please use outcome-specific study coordinates/codes and audit the same condition for the U and q trajectories.

tests/test_gp_orthogonal_anchor.py
```diff
@@ -0,0 +1,63 @@
...
+
+import vocab_growth.data_utils as vocab_data_utils
+from vocab_growth.models import common_univariate_re as cur
+from vocab_growth.models.common import ModelFitContext
+from vocab_growth.models.definitions import VG11
+
+
+@pytest.fixture
+def vg11_model(tmp_path, monkeypatch):
+    if not os.path.exists(vocab_data_utils.VOCABULARY_DATA_PATH):
```
frankbuckley, line 37: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex / GPT-5).

**[P2] This regression guard is skipped in CI.**

The fixture skips without `data/vocabulary.duckdb`, but the workflow runs `pytest` before `scripts/prepare_data.py`. The green CI jobs therefore do not exercise this test or the new assertions in `test_study_re_parameterisation.py`. Prefer a small synthetic PyMC unit test (the existing `test_gp_utils.py` scaffolding is sufficient), or run these specific model-construction tests after data preparation.

notes/202607191614-full-refit-rep-hightune-run.md
```diff
@@ -0,0 +1,114 @@
...
+- vg13 (TD joint young): fail at rep on dispersion + 28 div (1.037) → refit hightune tune 12000 / draws 8000 / **ta 0.99**.
+- vg11 (TD spoken, 16k obs): fitting at plain rep now; if it fails on trend/GP → refit hightune tune 12000 / **draws 6000** (memory) / ta 0.97.
+- Run the hightune refits **sequentially** (memory: vg11 16k + vg13 joint are heavy; never stack).
+
+**Then:** Phase B (comparisons + render docs/report + docs/comparison) → Phase C (upload to dseresearch/public; needs `az login`).
+
+**Provenance note:** published set will span commits 3e6f61d (baselines, code-identical) + b563586 (reparam) + the branch HEAD after this note. All docs/reparam commits are code-equivalent for each model's engine.
+
+**Infra lessons captured:** (1) never write into the working tree during a reporting run (dirties in-flight manifests); (2) launch long runs under `systemd-run` (system scope), not bare `setsid`/`nohup` (logind `KillUserProcesses` kills the latter on logout); (3) the hierarchical TD hightune fits are memory-heavy — draws 12000 OOMs on 251 GB; keep draws ≤ 8000 and never run two heavy fits concurrently.
```
frankbuckley, line 114: > [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex / GPT-5).

**[P2] Extend the durable run record through the final change and validation.**

This note ends with the TD refits still outstanding and does not record the orthogonal-GP commit or the VG10/VG15 improvements quoted in the PR description. Because the PR cites this file as the full run record, please append the final commit/configuration, convergence results, remaining failures, and the evidence supporting the reported before/after metrics.

frankbuckley (commented):
> [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex / GPT-5).

I found four substantive statistical/model-design issues and two evidence gaps. Details are inline. I would not merge until the P1 items are resolved, or explicitly reframed as intentional model revisions with updated prior justification, provenance, and validation.

src/vocab_growth/models/gp_utils.py
frankbuckley: > [!NOTE]

Agreed — this is a real defect, confirmed in the code. The projection statistics (`z.mean()`, `g_unit.mean()`, and the `zc·g` slope) are computed over the full `X_all_z` stack, which is observations ⧺ plot points ⧺ query ages ⧺ the anchor row (`common_univariate_re.py:316-318, 335`). The projected `g_unit` at the observed rows then feeds the likelihood, so changing `n_plot`, `ages_query`, or the anchor placement moves the posterior. Inference must not depend on the reporting grid.

Planned fix: estimate the projection on a **fixed model design** — the training (observed) ages — compute the mean and linear-projection coefficients there once, and apply those fixed coefficients to every evaluation row (obs, plot, query, anchor). Add an invariance test asserting that changing only the plot/query grids leaves the observed-row latent trajectory bit-for-bit unchanged.

This changes the likelihood, so the anchored models (vg10/vg11/vg12/vg13/vg15) will be re-fit; I'm holding the blob upload until they are.

src/vocab_growth/models/gp_utils.py
frankbuckley: > [!NOTE]

Agreed on both points. The new block never indexes `g_unit[anchor_idx]`, so `g` no longer passes through zero at the reference age — which directly contradicts the `anchor_g*_at_ref` contract stated in the docstrings (`common_univariate_re.py:212-213`, `common_joint_modality.py:714`), the `definitions.py:134` field comment, and the build output.

Planned fix: **preserve the point anchor and additionally remove the linear tilt**, i.e.
`g = g − g[anchor] − β·(z − z[anchor])`
with `β` estimated on the fixed training design (per the sibling comment). This keeps `g[anchor] = 0` (contract intact) while decoupling the GP's linear direction from `slope`. For `tent_and_gp` you're right that `[1, z]` is too small a basis — the three-anchor piecewise-linear mean spans more — so there the GP will be projected against the **actual tent basis** (the segment/anchor design columns), not just `[1, z]`. The docstrings/field contract and build output will be re-verified against the restored behaviour.

src/vocab_growth/models/common_univariate_re.py
frankbuckley: > [!NOTE]

Agreed — the "prior-preserving" framing is wrong and I'll correct it. `ZeroSumNormal(1)` has covariance `I − J/K`: marginal variance `1 − 1/K` (a 25% reduction at K=4) and off-diagonal `−1/K`, so the study offsets are both shrunk and negatively correlated, and `tau` unchanged does not restore the original scale.

Concrete proposal (not a straight revert — restoring the removed group-mean DOF would reintroduce the very intercept/RE-mean ridge this change exists to remove):

1. **Keep the sum-to-zero constraint** — it is the mechanism that removes the non-identified common-shift direction.
2. **Rescale to `ZeroSumNormal(sigma = sqrt(K/(K−1)))`** (equivalently bump `tau` by the same factor) so the marginal per-study prior is `Var(delta_k) = tau²` again, matching the original scale. Only the mean DOF and the inherent negative correlation remain — and those are exactly the intended constraint, not recoverable by construction.
3. **Relabel this as an intentional identifiability/prior change** in the PR body, model docs, and provenance — not a prior-preserving reparameterisation — and redo the prior-predictive / prior-sensitivity checks under the corrected prior.

Net effect: the marginal study-effect shrinkage scale is preserved, and only the non-identified mean direction is removed. Applied consistently across `common_univariate_re`, `common_bivariate_re`, `common_joint_modality`. This shifts the numbers slightly, so the RE models are re-fit as part of the same pass.

src/vocab_growth/models/common_joint_modality.py
frankbuckley: > [!NOTE]

Agreed. `delta_sign` is a `ZeroSumNormal` over the full `study_id` (`common_joint_modality.py:894`), but only a subset of studies contribute signing observations. The uninformed studies' `z_sign` values are pinned only by the constraint + prior and can absorb a common shift among the informed studies, so the global zero-sum does **not** guarantee that the sign-informed effects themselves sum to zero — it need not remove the sign-intercept ridge and it couples the informed effects through nuisance ones.

Planned fix: introduce an **outcome-specific `study_sign_id` coordinate** covering only the studies that contribute signing data, put the `ZeroSumNormal` over that, and map to observations via sign-specific study codes. I'll audit the U and q trajectories under the same test (I expect both are informed by all retained studies, in which case the global coordinate is correct for them, but I'll confirm rather than assume). Re-fit of the signing model follows.

tests/test_gp_orthogonal_anchor.py
frankbuckley: > [!NOTE]

Agreed — confirmed: CI runs `pytest` (`ci.yml:74`) **before** `scripts/prepare_data.py` (`ci.yml:77`), so both new guards (this file and the new assertions in `test_study_re_parameterisation.py`) skip for want of `data/vocabulary.duckdb` and never actually run.

Planned fix: rewrite them as **synthetic PyMC model-construction unit tests** with no DuckDB dependency (following the existing `test_gp_utils.py` scaffolding), so they execute in CI. The orthogonality/anchor and sum-to-zero invariants can all be asserted on a small hand-built design matrix.

notes/202607191614-full-refit-rep-hightune-run.md
frankbuckley: > [!NOTE]

Agreed. I'll extend the run record through the final change and validation: the orthogonal-GP commit and its correction, the final per-model commit/configuration, the convergence results (R-hat, min ESS, divergences, BFMI), the remaining caveats, and the provenance of the VG10/VG15 before/after numbers quoted in the PR description. Since the corrected fixes above require re-fitting, these figures will be regenerated from the re-fit and the note updated to match — the current numbers are from the parameterisation under review.

frankbuckley:
> [!NOTE]

Thanks — this is a thorough review and I accept all six points; I verified each against the code before responding. Summary of disposition (details inline):

- **P1-A / P1-B (GP anchor)** and **P1-D (sign zero-sum)** are behaviour-changing — they alter the likelihood — so the current reporting-quality fits for the anchored models (vg10/vg11/vg12/vg13/vg15) and the signing model are not trustworthy as they stand. These need re-implementation **and re-fitting**, and I'm **holding the blob upload** until the corrected fits are clean.
- **P1-C** will be resolved by keeping the identifiability constraint but rescaling to `ZeroSumNormal(sqrt(K/(K−1)))` so the marginal study-effect scale is preserved, and relabelling it honestly as an intentional prior change with fresh prior-predictive checks (inline).
- **P2-A / P2-B** are cheap: make the guard tests run in CI (synthetic, no DuckDB) and extend the run record through final validation.

I'll push the corrected fixes + re-fit and re-request review rather than merge as-is. Please flag if you'd prefer any of the P1s reframed as intentional model revisions instead of corrected — my current plan treats A/B/D as bugs and C as an intentional (now correctly documented) constraint.

frankbuckley:
For reference - a reminder as to why we are now seeing fit issues in these models:

> So why the ridges surfaced this run, when they hadn't before, is a combination of three things — and only the first is a model change:
>
> 1. The new subject-RE layer ([#164](https://github.com/dseinternational/vocabulary-growth/issues/164)). A per-child intercept over the narrow young-TD age window, with many children, adds variance that trades off against the population level, the dispersion kappa(z), and — through the shared intercept — the study-RE mean. It didn't create the study-RE-vs-intercept ridge (that one is structural and has been latent since #43), but it tightened the geometry enough to push several models over the R-hat gate.
> 
> 2. This is the first full-quality refit of these models. At dev/test config (few draws, low tune) a slow-mixing ridge direction just looks like noise; at rep/rep-hightune (12k+ tune, 8k+ draws) the sampler actually explores it and R-hat/ESS report the non-identifiability. Heavier tuning exposes the ridges rather than curing them — which is exactly why vg12 got worse at hightune.
> 
> 3. The TD data itself. Few studies (small K → the study-RE-mean ridge bites hard) and a narrow age window (the trend/GP-tilt and dispersion floor/slope directions are poorly separated) make all three ridges tighter for the TD models specifically, which is why the DS models mostly passed while the TD ones didn't.

frankbuckley:
> [!NOTE]

**Corrected fixes pushed** (`23f5ba1`, `71e46b4`) — all four P1s and both P2s addressed. Summary of what changed versus the reviewed revision:

- **P1-A / P1-B (GP anchor).** The projection coefficients are now fitted on the observed rows only (new `n_obs` argument) and applied to every row, so `n_plot` / `ages_query` no longer touch the observed-row likelihood; the residual is then pinned to zero at the reference-age row, restoring the `anchor_g*_at_ref` point-anchor contract. The nuisance basis is mean-specific — `[1, z]` for `trend_and_gp`, `[1]` for the free-intercept mean, and the three tent hats for `tent_and_gp` (the larger basis you flagged).
- **P1-C.** Kept the constraint (it is what removes the ridge) but rescaled `ZeroSumNormal` sigma by `sqrt(K/(K-1))` so each marginal prior variance stays `tau²`, and relabelled it in code/PR as an intentional identifiability constraint rather than prior-preserving.
- **P1-D.** `delta_sign` is now zero-summed over the sign-informed studies only, scattered into full `study_id` with 0 elsewhere. Verified on VG15 (5 of 12 studies active). U and q audited and kept global.
- **P2-A.** New data-free synthetic tests in `tests/test_gp_anchor_orthogonalisation.py` (anchor-zero, observed-row orthogonality, reporting-grid invariance, intercept/tent basis, RE-rescale marginal variance) — these run in CI; the integration guards were updated to the new contract. **P2-B.** Run note extended through the review and the corrected fixes.

Full suite green, ruff/spellcheck/prettier clean. Reporting-quality refits of the affected models are being regenerated on this code; **the blob upload stays held** until they are clean, and the superseded whole-grid `vg10`/`vg15` metrics will be replaced with numbers from the corrected fits. Re-requesting review once the refits land.

PR #180, opened 2026-07-25
feat(report): surface the posterior-predictive calibration evidence (#163)

frankbuckley:
> [!NOTE]

## Summary

Closes the quantitative-calibration half of #163's acceptance item "add quantitative posterior-predictive calibration plus simulation/parameter-recovery checks for the preferred models". The parameter-recovery half landed in #177.

**No fitting is involved.** Every fit has written `posterior_predictive_calibration.csv` since #164, and no report has ever read it — `grep -rl calibration docs/**/*.qmd` returned nothing. The evidence was computed, synced into the figure cache, and invisible. This surfaces it.

## What is reported, and where

Two places, both generated from the written tables so no number is hand-entered:

- **The workflow chapter** gains a `Quantitative predictive calibration` section: what is measured (per-outcome empirical coverage at the reported widths, mid-PIT mean/variance/extreme rate, mean error, pooled and in 12-month age bands), how to read it, and a cross-model table of every model of record.
- **All 15 per-model reports** gain a `Predictive calibration` section with their own per-outcome table and an age-banded breakdown, placed next to the posterior-predictive overlays it quantifies.

Both routes go through shared helpers in `models/calibration.py`, so the selection and the interpretation cannot drift between the report and a model's own page.

## Two things the prose is deliberately careful about

**These are in-sample diagnostics.** The tempting reading is that coverage above nominal with PIT variance below $1/12$ shows the reported intervals are conservative. That is not sound: the replications condition on parameters the same observations informed, so an under-dispersed PIT and coverage above nominal are *expected* even for a well-specified model. The chapter states this and points to the LOO/ELPD comparison as the out-of-sample counterpart, rather than letting an artefact of in-sample checking read as reassurance about a new child. For a report written partly for a clinical audience that distinction matters.

**No narrative about the current numbers.** The cached tables are all from the paused pre-#176 run — nine models, dated 16–18 July — so any pattern in them is superseded. The prose gives the reading rules and lets the generated table speak; the numbers arrive with the next figure sync. This is the same discipline `notes/202607251055-combined-sensitivity-recovery-run-plan.md` asks for.

## An inconsistency fixed on the way

Calibration was tabulated at (0.50, 0.80, **0.90**) while every credible interval in the report is **89%** — so the report could not have quoted predictive coverage at its own reported width. The tabulated levels are now tied to `vocab_growth.intervals` (0.50 / 0.89).

Tables written before this change carry a 0.90 outer level, so the reporting helpers **resolve the level from the table and label what they found**. An older fit's table stays readable and correctly labelled rather than being silently mislabelled as 89%, and the cross-model table carries a `Nominal` column so a mixed cache is honest. Both cases are covered by tests.

## Validation

Verified by **rendering**, not only by unit tests:

- The workflow chapter builds with the cross-model table populated for the nine models that have a cached table, and the other six absent without error (`show_or_pending` handles a partial cache).
- A per-model section renders both its tables with the nominal level stated and age bands ordered numerically — `[108, 120)` after `[96, 108)`, which a lexical sort would have put second and made the age trend unreadable.
- Full suite green (206 tests), `ruff`, `cspell` and Prettier clean.

New tests cover: the tabulated levels following the reporting convention; level resolution for both a 0.90 and a 0.89 table; numeric age-band ordering; display labels not leaking internal column names; and the per-model section explaining an absent table rather than rendering empty.

The one existing test that asserted the old levels now asserts against the convention rather than literals, so it cannot drift from `vocab_growth.intervals` again.

## Noted, not fixed

Rendering surfaces one pre-existing warning: `@sec-appendix-b1` is dangling because Appendix B is still a five-line `TODO` stub, while the workflow chapter says "the per-model evidence is in Appendix B". That sentence is now doubly stale, since the calibration evidence lives in the per-model reports. Worth its own issue.

Refs #163.

### DIFF
--- a/docs/models/vg01/index.qmd
+++ b/docs/models/vg01/index.qmd
@@ -222,9 +222,7 @@ Lower $\kappa$ values indicate greater overdispersion (greater between-child var
 
 ## Predictive calibration
 
-Quantitative counterpart to the posterior predictive checks above: how often each
-predictive interval contains the observation it predicts, and where the
-observations sit within their predictive distributions.
+Quantitative counterpart to the posterior predictive checks above: how often each predictive interval contains the observation it predicts, and where the observations sit within their predictive distributions.
 
 ```{python}
 #| echo: false
--- a/docs/models/vg01/index.qmd
+++ b/docs/models/vg01/index.qmd
@@ -219,3 +219,17 @@ The posterior distribution of the derivative of expected spoken words with respe
 Lower $\kappa$ values indicate greater overdispersion (greater between-child variability relative to the mean), while higher $\kappa$ values indicate less overdispersion (less between-child variability relative to the mean). The plot shows how the estimated dispersion parameter $\kappa$ changes with age, along with uncertainty intervals.
 
 ![posterior_kappa](posterior_kappa.png){#fig-posterior-kappa .lightbox fig-align="left"}
+
+## Predictive calibration
+
+Quantitative counterpart to the posterior predictive checks above: how often each
+predictive interval contains the observation it predicts, and where the
+observations sit within their predictive distributions.
+
+```{python}
+#| echo: false
+#| output: asis
+from vocab_growth.models.calibration import render_calibration_section
+
+render_calibration_section()
+```
--- a/docs/models/vg02/index.qmd
+++ b/docs/models/vg02/index.qmd
@@ -222,9 +222,7 @@ Lower $\kappa$ values indicate greater overdispersion (greater between-child var
 
 ## Predictive calibration
 
-Quantitative counterpart to the posterior predictive checks above: how often each
-predictive interval contains the observation it predicts, and where the
-observations sit within their predictive distributions.
+Quantitative counterpart to the posterior predictive checks above: how often each predictive interval contains the observation it predicts, and where the observations sit within their predictive distributions.
 
 ```{python}
 #| echo: false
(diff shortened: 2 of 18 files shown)


frankbuckley:
> [!NOTE]

Thanks — three distinct findings across the 17 comments, and all three were right. Fixed in `7974c06`.

**1. The in-sample caveat missing from the per-model text** (`calibration.py`) — the substantive one, and the catch I'm most glad of. The section said coverage above nominal with PIT variance below 1/12 means the predictive distribution is wider than the data warrant, and stopped there. The workflow chapter is careful about this, but a model's own page is read on its own, and in that context the reading is unsupported. The rendered text now states that these are in-sample checks, that some over-coverage and an under-dispersed PIT are expected even for a well-specified model, that the direction is therefore *not* evidence the reported intervals are conservative for a new child, and that the out-of-sample counterpart is the LOO/ELPD comparison. A new test asserts the caveat is present so it cannot be dropped again — the guarantee I should have written the first time, given this is exactly the misreading the chapter was drafted to prevent.

**2. Hard-wrapped paragraph in all 15 per-model reports** — correct, and my error. The new paragraph wrapped at ~80 characters while the surrounding prose in those files runs to 774 characters on a single line. `AGENTS.md` is explicit that prose is written as one continuous line, and with Prettier's `proseWrap: preserve` it would have stayed wrapped and rendered as mid-sentence breaks. It came from a heredoc that wrapped. Now a single line in all 15.

**3. Unused `nominal_level` import** in the chapter cell — dropped. `overall_calibration` already returns the level that is displayed, so the import only muddied which helper decides it.

Re-rendered both routes after the change: the workflow chapter builds with the cross-model table populated for the nine models that currently have a cached table, and the per-model section renders the caveat plus both tables. Full suite green, `ruff` / `cspell` / Prettier clean.

The one pre-existing warning I have not touched is the dangling `@sec-appendix-b1` crossref — Appendix B is still a five-line `TODO` stub, which is out of scope here and worth its own issue.

PR #183, opened 2026-07-26
docs: item difficulty, re-billed on Rasch sufficiency; fix(data): audit the Edgin subset and register the inverse sensitivity

frankbuckley:
> [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Fable 5).

> [!IMPORTANT]
> **Scope grew, and the framing was corrected. Read this section before the original description below.** The PR began as two working notes about item exchangeability. It now also carries a systematic data audit of the `us_01` (Edgin) subset with four defect classes masked by default, a registered sensitivity that reinstates them, a pre-specification for the differential-item-functioning test, and a CI fix. The original description is preserved beneath, because two review rounds are threaded against it — but several of its figures are superseded, and each is flagged there.
>
> **The two notes this PR began with no longer exist as separate files.** After the framing correction they carried three strata of appended corrections each, and the study owner judged — correctly — that the trail obscured the position. They are consolidated into [`notes/202607261540-item-difficulty-and-the-aggregate-likelihood.md`](https://github.com/dseinternational/vocabulary-growth/blob/docs/item-difficulty-proposals/notes/202607261540-item-difficulty-and-the-aggregate-likelihood.md), which presents the corrected argument in its final form with a corrections log (§15); the deleted files' full history is in git (`git log --follow`) and in this PR's threads. A new [`notes/README.md`](https://github.com/dseinternational/vocabulary-growth/blob/docs/item-difficulty-proposals/notes/README.md) indexes every note with its status. The filenames used as headings in the preserved description below refer to the deleted files.

## The framing was wrong, and the study owner caught it

The original description leads with "the assumption is false". True, and demonstrated on a test that fits no model — but it was the wrong thing to lead with, because it implies the likelihood is compromised.

Under a Rasch-type item model — items of arbitrary, heterogeneous difficulty but **equal discrimination** — the total score is a **sufficient statistic** for ability. Writing the log-likelihood as `theta*T - sum_j y_j d_j - sum_j log(1+exp(theta-d_j))`, the term that depends on *which* items were passed carries no `theta`. So modelling sum scores discards no information about ability, heterogeneous difficulty costs nothing on that axis, and item heterogeneity can reach the model through exactly one route: **the distribution of the total**. There is no other channel to look for. That is the principle the count-variance section computed without stating.

Measured against VG10's own fitted `kappa` rather than assumed:

| model / outcome  | fitted `kappa` | variance inflation | kernel's share of total variance | worst-case effect on total SD |
| ---------------- | -------------- | ------------------ | -------------------------------- | ----------------------------- |
| VG10 understood  | 5.3–44.0       | 19×–129×           | **0.77%–5.27%**                  | **1.06%**                     |
| VG10 spoken      | 5.8–6.0        | 117×–120×          | 0.83%–0.86%                      | 0.17%                         |

Between-child dispersion dominates the kernel by one to two orders of magnitude. **So the Beta-Binomial is adequate for the total, and item exchangeability is not a live threat to any estimate.**

Two corrections fall out, pulling opposite ways. The kernel misspecification is **larger** than the notes said: they took the ~1.8-logit gap between the outer checklists as a proxy for the item-level difficulty spread, which understates it, and at a realistic 2-logit spread the true conditional variance is ~60% of Binomial, not 91%. Also worth recording, because it is structural: the Beta-Binomial's variance is `N*p*(1-p)*(N+kappa)/(kappa+1)`, which is `>= N*p*(1-p)` for every `kappa`, so the family has the Binomial as a **floor** and cannot represent underdispersion at all — `kappa` absorbs it. And yet the exposure is now pinned by fitted output rather than bounded by assumption, and the product is still negligible. **The "documentation only, no refit" conclusion was right, for a reason the notes had not articulated.**

### What this changes, and what it does not

Nothing is dropped. The billing changes:

- **The exchangeability note's title over-claims** and now says so at the head. What survives is the **estimand channel** (`q` is a composition-weighted average, not the item-level probability the report defines) and the **level-scale channel** — which was never an exchangeability problem, being true even if every word were equally hard.
- **The proposals note states there is no fit-based case for any proposal**, and none should be argued on those grounds. **C** addresses a real estimand defect. **A and B are interpretive, not corrective** — B's difficulty-mixed link shifts the implied `kappa` by ~10%, a change to what `kappa` *means*, not to how well anything fits, and not a shift these data could detect.
- **The Route 1 pre-specification states why the work is warranted.** Two reasons, neither touched by sufficiency: the hypothesis *is* item-level, since "same order" is a claim about the difficulty vector, and sufficiency is precisely the theorem proving a sum score cannot test it; and one published contrast (`Delta q` at matched comprehension) depends on composition.
- **`methods-models.qmd`** now separates the two Binomial assumptions it had presented the Beta-Binomial as jointly remedying, states that item exchangeability survives and is false, and quantifies why that does not threaten the estimates.
- **One threat is added rather than removed.** Rasch sufficiency requires equal discrimination *exactly*, and CDI items do not have it. That does not revive the exchangeability concern — it lives in the same ~1% budget — but a 1PL fitted to items with varying discrimination can present as differential item functioning that is not there, and the two populations occupy different ability ranges by construction. A **2PL sensitivity is now registered in the pre-specification** as non-optional, with the 2PL governing on disagreement.

## Data: the `us_01` (Edgin) subset carries four defect classes

Chasing one residual — a maximum spoken count of 152 words at 11 months against a Berglund median near zero at 12 — showed that finding defects by hand was the wrong method. `scripts/audit_edgin_subset.py` enumerates every class the aggregate data can expose. **32 of 196 administrations (16%), 31 of 119 children.** Two checks came back clean, worth recording: no count exceeds its form's ceiling, and no form was administered outside its age band.

- **Outcome-column duplication (8).** Production within 25% of comprehension at 11–18 months. The ratio threshold is set from the largest gap in the measured distribution (0.86 → 0.55, width 0.306) rather than chosen; an earlier 0.9 cut ran through the middle of the cluster and missed two records, one of them the 152-word case.
- **Saturation at the form maximum (19).** Thirteen fall inside a run of **21 consecutive records in identifier order, every one at exactly 680 words**, ages 24–30, crossing into the autism and unlabelled groups of the same source cohort, with no other administration for any of the 21 children. Placing the 27 ceiling records at random among the cohort's 235, `P(any run >= 21) = 1.3e-22`, computed exactly. Age-matched: 13 of 37 DS records at 24–30 months sit at the ceiling (35%) against 0.47% of 5,135 typically-developing records, while the median of those same 37 is 35 words.
- **Longitudinal collapse (12).** Vocabulary does not shrink, so a count several-fold above the same child's later count is unambiguous — seven of eight young ceiling records are contradicted 27- to 223-fold. The rule needs a floor (else 5 words falling to 1 fires) and an age scope (at older ages a decline can follow a form change); two such records in other sources are left for separate investigation.
- **One administration recorded twice**, identically, which would double-weight it and make a single-visit child look like repeated measures.

Every exclusion falls in `us_01`. Pool 1,219 → 1,218 rows; understood 680 → 671; spoken 1,145 → 1,114. **Maximum spoken at 8–14 months falls from 152 to 12.** Each rule has an `include_*` flag.

**Three corrections to my own audit are recorded in the note rather than quietly fixed.** I described the ceiling block as "genuine right-censoring, not error" — wrong; they are invalid values. I described it as "a contiguous block of consecutive child identifiers" with "an 81-identifier gap" — both false: 22 integers inside the window are absent from Wordbank entirely, and the next identifier present is an autism-group child also at exactly 680. And I quoted a typically-developing reference of "1,469 children aged 16–19, maximum 643" — that took the audit's *Words & Gestures* count and attached it to a *Words & Sentences* claim; the correct figures are 0.00% of 2,489 with a maximum of 661.

### Provenance, and why the source cannot settle it

The study owner established with the source author that **the original data files no longer exist**. Checking what `us_01` actually is before accepting that turned out to matter: unlike every other study, `us_01` is not built from a study file at all — it is derived in the view as the `dataset_name = 'Edgin'` Down syndrome subset of the **public Wordbank by-child export** (196 administrations, 119 children), and it has no codebook. A tracked `data/vocab_data_us_01.csv` does exist, but nothing reads it — an orphaned 66-row Words & Gestures extract on a different identifier scheme; my first write-up asserted the file's absence without checking, and §13.1 of the note now carries the correction and the double-count trap it leaves. So the loss is narrower than it looked (what is gone is the ability to *explain* and *repair*, not to *detect*), and the available evidence is wider: the cohort's pre-term, autism and unlabelled groups passed through the same preparation and are internal controls the `us_01` filter had been hiding. Pre-term shows no ceiling saturation at all, so the defect is not a whole-file transformation.

It also surfaced a contamination the audit had missed: **two Edgin rows satisfy the typically-developing filter, and one is a 680-word record inside the flagged run** — sitting in the reference pool the exclusions are benchmarked against. Edgin is the only clinical cohort in the export that leaks this way (0.5% of its rows, against ≥10% for every other contributing dataset), so it is now barred from the pool. Two rows of 15,379; no estimate moves. Done because a record excluded as invalid on one side of a comparison cannot remain inside the other.

## The inverse sensitivity, registered

The `us01-ceiling-excluded` variants are **retired**: they excluded records now masked by default, so they could not fail, and a registered check that cannot fail is worse than no check. They are replaced by the inverse — **`us01-implausible-reinstated`** for VG10 and VG15 — which puts the masked production counts back and refits. Since the exclusion can never be confirmed at source, this pair is the only evidence a reader has for whether the headline trajectories depend on our judgement, so it is not optional in a published refit.

Verified by running both engines' preparation, not just the registry: VG10 spoken 1,114 → 1,136, VG15 marginal spoken 947 → 969. That is **22, not 30**, because 8 of the masked administrations are also caught by the duplicated-outcome rule, which stays active with its own flag.

Three ways this could have shipped looking fine and doing nothing, each closed: `load_data` had no way to pass the `include_*` flags through to `load_combined_data`, so a definition field alone would have changed the config name and nothing else (it now forwards all three, and **raises** for the typically-developing population, where these defect classes do not exist); each engine prints the reinstated count, derived by differencing the loader paths so it cannot drift from the rule, and a run reading 0 is a failure to investigate; and the first implementation read `definition.max_age_months`, which `JointModelDefinition` does not define, so **VG15 raised `AttributeError` the moment the engine ran** while every unit test passed — `tests/test_implausible_production_sensitivity.py` exercises both engines' preparation and asserts the prepared frames differ by exactly the reported count.

## CI

CI had been **red since the first Edgin commit**, and I had not checked it — two tests opened the real DuckDB, which is a gitignored build artefact absent on the runners, and passed locally for that reason. They now carry a `requires_real_db` marker, and **CI builds the database before pytest** instead of after: it costs ~300 ms, and it means the assertions pinning the study-level inclusion counts quoted in the notes and the report actually run there rather than skipping. A suite that skips them reports green while the counts drift. Verified by reproducing the CI condition locally with the database moved aside.

## Corrected figures in the original description below

- **"Within-child variance is only 0.6–4.6% of total count variance"** — close, slightly low at the top: measured against VG10 it is 0.77%–5.27%. The associated "effect on `kappa` is about 1%" stands.
- **"Six WG administrations are excluded"** in the pre-specification — now **eight**, after the ratio threshold was set from the measured gap. The Route 1 sample, recomputed through the loaders rather than by hand, is **78 WG comprehension administrations from 50 children** and **165 production administrations from 102 children**, not 81/50 and 190/116.
- **VG07 figures were quoted as "the fitted models"** in a draft of this work. VG07 is not a model of record and its `kappa` is far lower than VG10's, understating the exposure threefold. All figures above are VG10's, and `scripts/verify_item_difficulty_notes.py` now checks them against its output so the mistake cannot recur silently.
- **The `us01-ceiling-excluded` sensitivities** referred to as registered are retired; `docs/runbooks/full-refit.md` would have instructed the operator to fit variants that now raise `KeyError`, and has been corrected.

## Refit status

**Every model of record is stale in substance** — the analysis frame changed three times (the ie_01 partial-administration mask, the Edgin masking, the TD-pool exclusion). No number in the report should be quoted until a full `rep` refit runs, plus the two new sensitivity fits. 326 tests pass; ruff, cspell and prettier clean.

---

# Original description (preserved; two review rounds are threaded against it)

Two companion working notes plus a verification script; no model-code changes. Branched off `main` after #182 merged. Revised in place through two review rounds — see **Revisions** at the end; every figure below reflects the corrected notes.

## `202607261008-challenging-item-exchangeability.md` — the findings

The Beta-Binomial relaxes between-observation homogeneity of `p` but leaves **within-observation item exchangeability** fully intact: conditional on a child's ability, every word of the 810-item inventory is treated as equally likely to be known. `methods-models.qmd:26` presents the Beta-Binomial as remedying both; it remedies only the second.

The assumption is false on a test that fits no model — with the scope stated precisely: the test rejects the equal-item _completion_ of the aggregate model (the item-level reading used whenever `kappa` or `q` is interpreted per word), not the distribution of the totals themselves. Under exchangeability a child knowing `T` of 810 words has a stratum count distributed Hypergeometric — pure combinatorics from the known checklist sizes:

| statistic                   | observed             | predicted |
| --------------------------- | -------------------- | --------- |
| RMS z, all strata           | 9.61                 | 1.0       |
| Checklist 1 excess positive | 37/44 (p = 5.3e-06)  | ~22/44    |
| Checklist 3 deficit negative | 42/44 (p = 1.1e-10) | ~22/44    |
| `p_C1 >= p_C2 >= p_C3`      | 79.5%                | 19.2%     |

None of 2000 exchangeability simulations reached the observed monotonicity rate.

**The note's main contribution is separating three concerns that had been travelling under one label.**

1. **The count-variance channel is negligible** — under local independence, and the natural intuition is backwards. By the exact identity `sum_i p_i(1-p_i) = N*pbar*(1-pbar) - N*Var_items(p)`, difficulty heterogeneity _reduces_ within-child variance. Within-child variance is only 0.6–4.6% of total count variance (the between-child term carries N² = 810²), so the effect on `kappa` is about 1%. Item dependence beyond ability is a separate, unquantified channel that the observation-level `kappa` partially absorbs.

2. **The consequential channel is not an exchangeability problem at all.** On a bounded proportion scale with a logit link, `kappa + 1 ~= 1/[p(1-p)*sigma^2]`, so a _constant_ latent between-child spread already implies a declining `kappa` — U-shaped, minimised at p = 0.5. `§sec-kappa` correctly labels the monotone decline an **assumption**, and that framing should be kept. But its parenthetical defence — that the bounded outcome means the decline "allows greater heterogeneity around the mean trajectory on the latent scale" — is backwards, twice over in VG10: the model already carries an explicit constant between-child latent spread (`tau_subj_u`, posterior mean 0.754) underneath the Beta step, so `kappa` there is residual, occasion-level dispersion; and reading the model's own fitted dispersion (`posterior_kappa_u.csv`):

   ```
   age    p      kappa   residual latent SD   total with tau_subj
    12  0.028    39.77         0.956                1.218
    30  0.227    25.66         0.462                0.884
    66  0.526    11.68         0.562                0.941
   ```

   On the `kappa + 1` scale the identity governs, the fitted decline over 12–66 months is 1.168 log units where a constant latent spread predicts 2.230 — a ratio of 0.52. And since `b_mag > 0` is imposed, the exponential-decline family cannot express the U-shape, so a positive fitted `b_mag` is by itself uninformative about latent divergence. Mitigating: the discussion and summary chapters and the signed results chapter are TODO stubs, and the substantive results chapter carries no dispersion prose, so no published text currently draws the "children fan out" inference; it appears as a recommendation in the earlier review note, and acting on that would be the error. The count-scale fan-out survives — posterior-predictive spread widens roughly five-fold from 12 to 66 months — so the `P(Y<=k)` columns and predictive intervals, not `posterior_kappa`, are the artefacts to lead with.

3. **The genuine exchangeability bite is on the estimand.** `q` is steeply difficulty-graded — the within-child log-odds gap between the outer checklists has median +2.53, an odds ratio of about **12.6** (n = 30, 27/30 positive, exact Wilcoxon p = 4.7e-07) — so `q` is a composition-weighted average over each child's own understood set, not the item-level probability `§sec-ratios` defines. Its level is not transportable, and it is acutely summary-dependent (the gradient roughly halves between a ratio-of-sums and a median-per-child summary). **Its rise survives.** A Kitagawa decomposition of the low-to-high comprehension split puts the composition component at −16.2% — composition _masks_ part of the gain there, negative in every specification tried; for the age gradient specifically the point estimates suggest attenuation and the intervals span zero, so no more than that is claimed.

**Not affected**, stated to forestall over-correction: mean trajectories and expected word counts wherever data constrain the fit (`E[y] = 810*p`), and the DS-vs-TD dispersion ratio (0.2–1.2%).

Magnitudes are graded honestly — existence and direction strong, magnitudes provisional at n < 50 — and gated behind a screening rule plus a codebook question, because 27 ie_01 records carry Checklist 1 counts above the nominal 120 (max 124), and the baseline wave's Checklist 1 field is unreliable (comprehension _falls_ between waves for 22 of 46 children while totals rise).

## `202607261048-incorporating-item-difficulty.md` — the proposals

Three changes, **none requiring a new likelihood family** (there are 27 `pm.BetaBinomial` construction sites across seven engine modules; none of these touches them):

- **A. Put the age variation on the latent scale.** `tau_subj_*` is a scalar `HalfNormal` in every engine that carries subject intercepts (the three RE engines behind all headline models) while `kappa(z)` carries all the age variation — the wrong way round for answering "do children fan out". `kappa` already has a single shared seam (`build_kappa_of_z`, 31 closure invocations), so this is cheap. Framed as the **first registered sensitivity**, not a demonstrated fix: scaling one per-child offset by `tau(age)` imposes perfect rank correlation across age, so random-slope and longitudinal-function variants stay on the table — and it needs a recovery run, because in the fitted frame 282 of 613 children (46%) are singletons, so `tau_subj` and `kappa` are already partly confounded.
- **B. A difficulty-mixed inverse link**, `p = sum_k w_k * logistic(f - d_k)`. The anchor-invariant flattening is `f(0.9) - f(0.5)`: 2.20 plain against 2.34 mixed, growing toward the ceiling — developmentally plausible, and it makes `f` interpretable as ability rather than the logit of a composition-dependent proportion. Whether the current link's ceiling approach is actually _wrong_ is an extrapolation claim only a refit with observed-scale predictive checks can demonstrate. The cost is more than the 35 `sigmoid` sites with no shared seam: the probability-specified anchors are converted with `logit` in `trend_and_gp` and the learning-rate artefacts hard-code the logistic derivative, so anchor re-derivation and downstream revalidation come with the link swap — with `d_k = 0` reproducing the current graph exactly.
- **C. Stratified means for both outcomes**, yielding a production propensity `h` that is composition-free _conditional on the calibration_, on which the cross-population contrast could be run instead of marginal `q`. One exactness caveat carried explicitly: with heterogeneous items `E[S | U = u]` is not exactly `u * p_S / p_U`, so keeping the nested `S | U ~ BetaBinomial(U, q)` structure under stratified means is a **moment-level approximation** — state it, score the stratum counts where they exist, or model `S` marginally.

**It also revises the earlier "full IRT is not close" conclusion** — conditional on the ingest audit. With item-level data reported for roughly 218 children with Down syndrome (~99 on the DSE inventory, 119 on the CDI) plus the Wordbank corpus, a Rasch calibration becomes feasible and the differential-item-functioning question becomes testable.

The note sets out that test. The "same order, slower" hypothesis — read as the cross-sectional measurement claim — decomposes onto two parameters: **"slower"** is the ability offset; **"same order"** is `r = sigma_DIF / sigma_d`, with population-specific variance share `r^2/(1+r^2)` and the item-pair rank-reversal probability as the estimand closest to "same order". Design points: identification needs **two** location constraints — the DIF zero-sum and the global Rasch location `sum d_j = 0` (the codebase already uses `ZeroSumNormal` for this pattern); the test targets the _spread_ `sigma_DIF`, whose ensemble identification is itself a hypothesis to verify by recovery and prior-sensitivity checks, with child effects for the repeated Edgin administrations; and item _classes_ give more power and interpretability than per-item DIF. Two linking routes in cost order — the **Wordbank Down syndrome subset (119 children, Edgin — already in this repository in aggregate as us_01) sits on the same CDI forms as the 35,025 typically-developing children, so it needs no item crosswalk at all** and is the cheapest first test.

Key architectural point running through all of it: item difficulty is _treated as_ a property of the **instrument** — the invariance assumption the DIF test checks — so a few hundred children measured at item level calibrate the inventory, and the calibrated inventory then serves the 1,219 aggregate observations conditionally: directly for the DSE-form rows, through linked item coverage for the shorter forms. Item-level data is not needed at fit time.

## Also

Adds `scripts/verify_item_difficulty_notes.py`, which recomputes the checkable figures in both notes from the raw CSVs and (where present) the fitted VG10 output — 60+ CLAIM-vs-COMPUTED checks, non-zero exit on mismatch; all pass on the current tree. Adds `Rasch`, `logits`, `reweighting`, `Haldane` and `Hermite` to the spellcheck allow list. ruff, cspell and prettier clean.

Two corrections to earlier working claims are recorded in note 1 §8 so they are not propagated: uk_01's 19 comprehension categories _do_ reconcile exactly with `understood` for all 29 complete rows; and the first circulated §3B kappa table used the wrong age standardisation — the corrected values _strengthen_ the conclusion. Two magnitudes remain from the supporting investigation's simulations and are flagged provisional until regenerated by committed code: the ≤1.1% likelihood-level effect on `kappa` and the 0.2–1.2% movement in the dispersion ratio.

## Revisions

- 1225bdd — corrections from an independent verification pass that recomputed every checkable number ([verification comment](https://github.com/dseinternational/vocabulary-growth/pull/183#issuecomment-5083097857)): the §3B age-standardisation error, `tau_subj_u` read alongside `kappa`, note 2's tables recomputed exactly with method and anchor stated, the exact Wilcoxon p, corrected seam counts, and the §6 attribution fix.
- 0ca26d8 — folds in the [review on this thread](https://github.com/dseinternational/vocabulary-growth/pull/183#issuecomment-5083074264) ([point-by-point response](https://github.com/dseinternational/vocabulary-growth/pull/183#issuecomment-5083163356)): the §2 scope clarification, the `kappa + 1` harmonisation, Proposal C's approximation label, the second DIF location constraint and variance-share correction, Proposal B's anchor and derivative costs, fitted-frame singleton counts, transportability scoping, the A1 reframe, and the committed verification script.


frankbuckley:
> [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

## Overall assessment

I recommend substantial revision before treating this PR as a settled scientific analysis or implementation roadmap. Its narrow empirical warning is valuable: the DSE inventory is decisively difficulty-ordered, fitted κ must not be presented as direct evidence of latent-scale “fan-out”, and q is inventory/composition weighted. Several downstream claims—especially Proposals A1, B and C and the Rasch/DIF interpretation—go beyond the evidence or are incorrect as written.

## What I independently verified

I reproduced the principal raw-data results from the committed ie_01 CSV: RMS z = 9.6128; 37/44 Checklist-1 excesses; 42/44 Checklist-3 deficits; 35/44 monotone profiles; the three checklist-specific q summaries; and the Kitagawa decomposition (+0.33993 total, +0.39497 within-stratum, −0.05504 composition, or −16.2%). These establish a strong difficulty ordering and a strong within-child production gradient. They do not establish every subsequent model claim.

## Major issues

1. **The hypergeometric result does not directly falsify the aggregate Beta-Binomial likelihood.** A count model assigns probabilities to the total T. It does not uniquely say which T words are known unless augmented with the conventional uniform-allocation item model. The hypergeometric test decisively rejects that item-level completion, but it is not a goodness-of-fit test of the observed-total distribution itself. The defensible conclusion is that the aggregate model is silent about composition and its usual equal-item generative interpretation is false—not that every aggregate model has thereby been falsified. See [the findings note, lines 17–39](https://github.com/dseinternational/vocabulary-growth/blob/ba157aee84475d794068fb172554c82152104830/notes/202607261008-challenging-item-exchangeability.md#L17-L39).

The related variance argument is conditional on independent items. Unequal fixed difficulties reduce Poisson-Binomial variance, but semantic clustering, prerequisite relations or other item dependence add covariance terms that can scale as N². “The count-variance channel can be set aside” should therefore be narrowed to difficulty heterogeneity under local independence.

2. **The correction to κ is right, but the claimed latent-spread profile is not established.** The report should not use decreasing κ as evidence that children fan out on a latent scale. The implementation already separates persistent subject effects from residual Beta-Binomial dispersion: tau_subj enters the latent predictor, while κ remains observation-level residual dispersion conditional on those effects ([implementation](https://github.com/dseinternational/vocabulary-growth/blob/b305e1f7755f8cab66278a507aab4006c6a6603e/src/vocab_growth/models/common_bivariate_re.py#L561-L591)).

The table claiming latent spread falls and then rises is only a delta-method transformation of posterior means. It ignores the existing subject-effect variance, treats Beta mixing as logistic-normal variation, omits posterior covariance and intervals, and derives a relationship for κ+1 but compares changes in κ. It establishes that a decline in κ can occur without growing latent spread; it does not estimate the actual developmental profile of latent spread.

Proposal A1 is consequently a useful sensitivity model, not a demonstrated fix. Multiplying one subject offset by tau(age) imposes perfect latent rank correlation across ages; random slopes or longitudinal random functions are plausible alternatives. Parameter recovery under A1 would test recoverability if A1 were true, not whether A1 is the right developmental structure.

There is also a concrete data-count error. Rebuilding the current data and applying the VG10 masking/filtering path gives **613 subjects, 282 singletons and 331 repeated subjects**, not 626/235/391. The PR’s figures count the age-valid raw view before removing 13 children with no surviving model outcome. Subject effects are also not present in “every engine”.

3. **Proposal B demonstrates a change of latent scale, not a better age trajectory.** For fixed latent f, difficulty mixing flattens the inverse link. But fitted f(age) is flexible, so both inverse links can represent essentially the same observed p(age). Showing that p = 0.9 occurs at f = 2.81 rather than 2.22 does not show that the current trajectory approaches the ceiling incorrectly. “Developmentally realistic” and “what the present link gets wrong” require refitted observed-scale predictive evidence, not a fixed-f comparison. See [Proposal B](https://github.com/dseinternational/vocabulary-growth/blob/ba157aee84475d794068fb172554c82152104830/notes/202607261048-incorporating-item-difficulty.md#L54-L71).

Nor is this a 35-site mechanical refactor. Current probability anchors are converted to latent anchors with logit(p) ([gp_utils.py](https://github.com/dseinternational/vocabulary-growth/blob/b305e1f7755f8cab66278a507aab4006c6a6603e/src/vocab_growth/models/gp_utils.py#L103-L123)). A mixed link requires its own inverse mapping or every anchor prior changes meaning. Learning-rate calculations also hard-code the logistic derivative p(1−p). All anchor priors, sensitivity variants, derivatives and extrapolation behaviour would need revalidation.

4. **Proposal C cannot exactly preserve the current nested Beta-Binomial likelihood.** The current model fits S given U as BetaBinomial(U, q, κS). With heterogeneous item probabilities, pS/pU is the conditional probability for a randomly sampled item, but generally E[S | U = u] is not u × pS/pU: conditioning on the observed total U changes which items are likely to comprise that total. Proposal C therefore cannot simultaneously claim an exact composition correction and “still a Beta-Binomial on totals” ([Proposal C](https://github.com/dseinternational/vocabulary-growth/blob/ba157aee84475d794068fb172554c82152104830/notes/202607261048-incorporating-item-difficulty.md#L73-L86)).

It needs observed stratum counts, a joint aggregate item-state likelihood/convolution, or an explicit statement that the retained Beta-Binomial is a moment-based approximation. Likewise, h would be a model-standardised propensity conditional on correctly calibrated and shared d/e, appropriate form weights, separability and measurement invariance—not unconditionally “composition-free”. Its existing independent q anchors would also cease to be valid because q would depend jointly on h, fU, d/e and form weights.

5. **The DIF proposal contains mathematical and identification errors.** The quantity σDIF/σd is an SD ratio, not a “fraction of difficulty variation”. If r = σDIF/σd, the corresponding fraction of total independent variance is r²/(1+r²). Nor does that ratio directly measure shared rank order; posterior rank correlations or item-pair reversal probabilities are needed.

Zero-summing delta_j resolves the group-shift alias but not the global Rasch location alias between theta_i and d_j; an additional constraint such as sum(d_j) = 0 is required. “Same order, slower” is also too causal: cross-sectional item data can test a shared item hierarchy plus an age-associated ability offset, not individual acquisition order or within-child learning speed.

The claim that σDIF will be well identified is a hypothesis requiring recovery, prior-sensitivity and boundary diagnostics. The Wordbank DS data comprise 196 administrations from 119 children, with substantial repetition; thousands of TD records make shared difficulties precise but do not guarantee precise DS DIF in a small, floor-heavy sample. Finally, if DIF is material, theta is not automatically a comparable fallback—it still requires invariant anchors, partial invariance or an explicit linking convention. See [the DIF proposal](https://github.com/dseinternational/vocabulary-growth/blob/ba157aee84475d794068fb172554c82152104830/notes/202607261048-incorporating-item-difficulty.md#L114-L147).

6. **Difficulty is not yet transportable across every pooled observation.** “Item difficulty is a property of the instrument” is itself the invariance assumption the DIF analysis proposes to test. Difficulty may vary by population, age, language, form, administration and response mode. The pooled analysis contains 396-, 416-, 675-, 680- and 810-item instruments ([current form inventory](https://github.com/dseinternational/vocabulary-growth/blob/b305e1f7755f8cab66278a507aab4006c6a6603e/src/vocab_growth/data_utils.py#L295-L304)); a DSE calibration cannot simply serve all 1,219 observations. Each form needs linked item coverage, or downstream analysis must be restricted to the calibrated inventory.

The headline “218 children” also combines different instruments. Only approximately 99 are asserted to have DSE-inventory data; CDI children do not calibrate DSE items without a crosswalk. No item-response extract, version/provenance manifest or coverage audit is included, and the 40-child source still has ages “to confirm”. The item-data audit must precede the claim that the previous feasibility constraint has been lifted.

7. **The correction to q should be narrower.** q = P(speak | understood) remains a legitimate conditional probability when defined over a randomly sampled word from the relevant inventory and understood set. It is an inventory-weighted aggregate, not an item-invariant production propensity. Such q values remain descriptively comparable, but differences do not isolate propensity at fixed item difficulty.

The Kitagawa point estimate suggests composition attenuates the low-to-high comprehension difference. It does not establish that composition conservatively biases the fitted **age** trend: the age-specific intervals span zero. The appropriate wording is “point estimates suggest attenuation; there is no evidence that composition explains the rise”. The report already includes the principal composition caveat ([current results limitation](https://github.com/dseinternational/vocabulary-growth/blob/b305e1f7755f8cab66278a507aab4006c6a6603e/docs/report/results-words-understood-spoken.qmd#L73-L78)); what remains insufficiently prominent is the 8–13-month age difference at matched comprehension.

8. **Several robustness claims are identities or unreproduced calculations.** E[Y] = 810p proves that p remains the model mean; it does not prove that fitted posterior trajectories are unchanged under a different likelihood, link or instrument model. I could reproduce the raw-data tables, but not the reported 1.1%, 13–16%, 2.5%, or 0.2–1.2% simulation effects: the PR includes no calculation script, seed-complete specification, fitted artefact or manifest. Those magnitudes should remain explicitly provisional until generated by committed code.

## Recommended corrected conclusion and sequence

The defensible conclusion is:

- The DSE inventory is decisively difficulty-ordered.
- The aggregate count models do not identify item composition, but this alone does not invalidate their total-count likelihoods or compel an immediate refit.
- κ must be described as residual probability-scale dispersion and cannot support a latent “children fan out” claim.
- q is inventory- and composition-dependent; the matched-comprehension comparison remains descriptive and confounded by composition and age.
- Item-level Rasch/DIF work is a worthwhile next study, conditional on a verified and versioned data ingest.
- A1/A2 and B/C should be competing registered sensitivities, not promoted according to the present sequence.

I would therefore: (1) correct the exchangeability, κ and q documentation now; (2) add reproducible calculation code and item-data provenance; (3) define a narrow measurement-invariance estimand, scope and practical equivalence threshold; (4) run Route 1 as a young-English-CDI feasibility study while auditing DSE items; and (5) compare A1/A2 and B/C as sensitivity models with calibration uncertainty, recovery and observed-scale predictive validation before changing a model of record.

frankbuckley:
> [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Fable 5).

I ran an independent verification pass over both notes — every checkable number recomputed from the raw CSVs (`vocab_data_ie_01.csv`, `vocab_data_uk_01.csv`, `vocab_data_merged.csv`), the fitted VG10/VG12 output, and the code tree. **The core findings hold and most figures reproduce exactly; the errors found are corrected in 1225bdd, and none weakens a conclusion — the two §3B corrections strengthen it.**

## Verified exactly (no change needed)

The §2 hypergeometric test in full (RMS z 9.61, mean z +8.40/+3.00/−9.01, 37/44 and 42/44 with sign-test p = 5.3e-06 and 1.1e-10, monotone 35/44 = 79.5%, and a fresh 2000-replicate null with mean 19.1% and no replicate reaching the observed rate). The §3A identity and shares. Every §3C summary: all three `q_k` rows, the Kitagawa decomposition to four decimals (+0.3399/+0.3950/−0.0550), and the weight shifts 0.422→0.210, 0.124→0.312. All §6 defect numbers (0.855→0.671, 252→323, 22/46 min −124, 27 records max 124, uk_01's 19 categories reconciling exactly for all 29 rows, 2 children with spoken/understood > 1, max 1.95). The headline Δq (+0.063 at N = 175, 89% CI [+0.021, +0.095]) against `ds_td_comprehension_q_at_U.csv`, and the 29.2 vs 16.8-month ages at 175 words from the two models' summaries. The counts: 1219 rows with age, 626 children, 235 singletons, 391 repeated; 680 = 739 understood obs − 59 ie_01 baseline rows masked by #182; native-only 259 obs / 194 children / four sources. The 35 `math.sigmoid` sites across exactly seven engines with no shared seam, and the single `build_kappa_of_z` seam. All quoted report passages (`methods-models.qmd` lines 26/47/111, the caveats stub, the `comparison.py` docstring, the earlier review note's §5). A bonus consistency check: `us_01` in this repository **is** the Edgin Wordbank subset — 119 children, 87 WG records with comprehension, 109 WS without — matching note 2's table exactly, so Route 1's aggregates are already in-house.

## Corrected in 1225bdd

1. **Note 1 §3B kappa table — age-standardisation error.** The tabled values implied z standardised on the understood subset (mean 34.6/SD 16.6) rather than the model frame's statistics (40.5/20.7); VG10's own `posterior_kappa_u.csv` gives 39.77/29.64/25.66/16.97/11.68 at those ages (the note had 8.94 at 66 months — 23% low). The fitted log decline becomes **1.225** against the constant-spread prediction 2.230 — ratio **0.55**, was 0.67 — so the "no evidence of latent fan-out" conclusion gets *stronger*. Recorded as a correction in §8 so the superseded figures don't propagate.
2. **Note 1 §3B now reads `tau_subj_u` = 0.754 alongside `kappa`.** VG10 carries an explicit constant between-child latent SD under the Beta step (same `diagnostics.csv`), so the implied-SD column is residual, occasion-level dispersion; a "total, with `tau_subj_u`" column is added (1.22 → 0.88 → 0.94 over 12–66 months — nearly flat after 30 months). This also ties §3B to the companion note's Proposal A, which is built on exactly this parameter.
3. **Note 1 §3C p-value.** The paired contrast reproduces only with a +0.5 continuity correction and both outer strata non-empty (that is where n = 30 comes from); with that construction the exact Wilcoxon p is **4.7e-07**, not 1.4e-06. The construction is now documented in §8.
4. **Note 2 §1 table recomputed.** The printed values matched neither the delta method nor exact integration under any single sigma, and implied a kappa inflation (6%–26%) inconsistent with note 1's "13%–16% roughly uniform". Recomputed exactly (Gauss–Hermite, sigma = 1, method now stated): plain 16.28/6.47/4.76, mixed 17.94/7.44/5.50 — inflation a fairly uniform 10%–13%, flattening ~3%, and both notes now say the same thing.
5. **Note 2 §3 table — anchor conflation.** The table's `d_k` turn out to be −logit(pooled stratum proportions), an anchor carrying weighted mean difficulty +0.44, so the "f for p = 0.5 rises 0 → +0.45" row was mostly the anchor convention. The anchor is now stated, an anchor-invariant row `f(0.9) − f(0.5)` (2.20 plain → 2.34 mixed) is added, "markedly" is retired in favour of the honest magnitude (+0.14 logits at p = 0.9, growing toward the ceiling), and two grid artifacts are fixed (peak dp/df is exactly 0.250; logit(0.9) = 2.20).
6. **Note 2 §6 attribution.** The `P(say | imitate)` finding is not in the companion note (its only "imitat" match is inside the word "limitation"); re-attributed to the supporting investigation and hedged with a quick pooled check of ie_01 (directionally consistent: 0.85/0.78/0.63 vs 0.68/0.56/0.63, but not clean within children at these n).
7. **Cross-note staleness.** Note 1's "full IRT … not feasible and not close" and "impossible retrospectively" are now cross-referenced to the companion note's §5 revision, so note 1 read alone no longer ends on a retracted conclusion.
8. **Smaller fixes.** The "results-\*.qmd are TODO stubs" claim (results-words-understood-spoken.qmd is substantive since #182 — though it contains no fan-out prose, so the mitigation stands); seam counts (27 `pm.BetaBinomial` construction sites, 31 kappa-closure invocations); `tau_subj` scoped to the three RE engines; 626/235 labelled as pooled-dataset counts; §5's mean-trajectory bullet gains the in-sample qualifier; §7 rec 3 extended to the `§sec-kappa` motivation sentence and the count-scale replacement message (predictive spread genuinely widens ~five-fold, so `P(Y<=k)` and predictive intervals carry the clinical message, not `posterior_kappa`); rec 6 softened to a reword (the docstring's claim is true in its narrow sense); a provenance note added for the external item-level datasets; `Haldane` and `Hermite` allowed in spellcheck.

## Superseded numbers in this PR description

The description above still shows the pre-correction figures; per the corrected notes: kappa at 66 months 8.94 → **11.68**, decline 1.484 → **1.225**, ratio 0.67 → **0.55**; Wilcoxon p 1.4e-06 → **4.7e-07** (OR ≈ 12.6); "f for p = 0.9 rises +2.22 → +2.81" → the anchor-invariant flattening is **2.20 → 2.34**; the §1 implied-kappa values and "five-fold" → **16.3 → 4.8, three-to-four-fold**; 30 BetaBinomial call sites → **27**; 24 closure invocations → **31**. Worth editing the description if you want it self-consistent.

cspell and prettier both pass after the changes.

frankbuckley:
> [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Fable 5).

Replying to the review in [this comment](https://github.com/dseinternational/vocabulary-growth/pull/183#issuecomment-5083074264), which was posted against ba157ae — my verification comment and commit 1225bdd landed about fifteen minutes later against the same revision, so on the central κ point (τ_subj on the latent predictor, κ as observation-level residual dispersion; "every engine" overstated) we converged independently; 1225bdd had already applied those. This comment records what I checked in the review's claims, what I have now adopted in 0ca26d8, and where I hold a different position.

## Independently verified from the review's claims

Its **fit-frame counts reproduce exactly**: rebuilding the age-valid frame and applying the #182 masking gives 613 children, 282 singletons (46%), 331 repeated — the review is right that the PR's 626/235/391 was the raw view, and the fitted-frame singleton share is materially worse for the τ_subj/κ confounding argument. Its **Proposal B cost claims verify in code**: the slope anchors are probability-specified and converted with `logit` inside `trend_and_gp` (so every anchor prior changes meaning under a mixed link), and `plot_expected_learning_rate` hard-codes `dE[Y]/dx = n·p·(1−p)·df/dx`. Its **point 7 premise verifies**: the composition limitation is already in `results-words-understood-spoken.qmd` (added in #182), so the report-side gap is narrower than note 1 first implied. And its **Proposal C mathematics is right**: with heterogeneous items, `E[S | U = u] ≠ u·p_S/p_U`, so the retained nested Beta-Binomial is a moment-level approximation, exact only under exchangeability.

## Adopted in 0ca26d8, by the review's numbering

**(1)** Note 1 now states that the hypergeometric test rejects the equal-item completion of the aggregate model — the item-level reading used whenever κ or q is interpreted per word — not the distribution of totals; and §3A is scoped to difficulty heterogeneity under local independence, with item dependence named as a separate unquantified channel that the observation-level κ partially absorbs. **(2)** The κ+1/κ mismatch is fixed — the decline comparison is now made on κ+1, the scale the identity governs (1.168 vs 2.230, ratio 0.52; 1.225 on κ itself), flagged as point arithmetic without posterior intervals; A1 is reframed as the first registered sensitivity with its perfect-rank-correlation fan shape named and random-slope/longitudinal alternatives kept on the table; and the singleton counts are restated for the fitted frame (282 of 613). **(3)** Proposal B's cost now includes anchor re-derivation and the hard-coded logistic derivative, and "gets it wrong at the top end" is re-stated as a prior-driven extrapolation claim that only a refit with observed-scale predictive checks can demonstrate. **(4)** Proposal C carries the exactness caveat, the conditional (not unconditional) composition-freeness of `h`, and the note that q anchor priors must be re-expressed once q is derived. **(5)** The DIF design gains the Rasch location constraint (`sum d_j = 0`) alongside the DIF zero-sum; the SD-ratio/variance-share correction (`r²/(1+r²)`); rank-reversal probability named as the estimand closest to "same order"; σ_DIF identification demoted to a hypothesis needing recovery and prior-sensitivity checks, with child effects required for the repeated Edgin administrations; a linking convention required before θ comparisons under material DIF; and "same order, slower" read as the cross-sectional measurement claim, not the developmental mechanism. **(6)** Transportability is scoped — the calibration serves non-DSE forms only through linked item coverage, "instrument property" is named as the invariance assumption §6 tests, the 218 headline is split into ~99 DSE / 119 CDI, and the feasibility revision is gated on the ingest audit. **(7)** The conservativeness claim is tightened to what §3C licenses — level-split attenuation consistent in sign across every specification; age-gradient point estimates suggestive with intervals spanning zero — and rec 4 now points at the missing piece, prominence for the 8–13-month age offset. **(8)** The repository now carries `scripts/verify_item_difficulty_notes.py`, which recomputes the checkable figures in both notes from the raw CSVs and (where present) fitted VG10 output with CLAIM-vs-COMPUTED output and a non-zero exit on mismatch; all checks pass on the current tree, and the two magnitudes it does not yet cover (the ≤1.1% κ effect and the 0.2%–1.2% dispersion-ratio movement) are flagged provisional in §8 until regenerated by committed code.

## Where I hold a different position

Three places, all narrow. First, I have kept the §3B consistency table rather than retreating to the purely negative claim: the quantitative comparison against the level-effect benchmark is precisely what licenses "the fitted decline furnishes no evidence of latent divergence", and it is presented as a consistency reading of point summaries — §8 states it is not an estimate of the latent-spread profile, which I take to be the review's actual requirement. Second, for the comprehension-level split I have kept the sign statement: attenuation is negative in every specification tried, and that verified consistency is informative beyond "no evidence" — the weaker wording is adopted for the age gradient, where it is what the intervals support. Third, on the overall assessment: with these revisions the notes' stated conclusions now match the review's own "defensible conclusion" list essentially point for point, and since the PR changes documentation and proposals rather than any model of record, I read the right disposition as merge-as-revised-working-notes, with the review's process additions — a pre-specified invariance estimand and practical-equivalence threshold, and observed-scale predictive validation before any promotion — carried into the sequence, as they now are.

One correction to the review where it reviewed a stale claim through no fault of its own: its point 8 could not reproduce the 13–16% / 2.5% figures because they were wrong — 1225bdd had already replaced them with the exact Gauss–Hermite values (a fairly uniform 10%–13% inflation; ~3% flattening), and those are now also enforced by the committed script.

PR #192, opened 2026-08-02
feat(models): parameterise kappa on two age anchors, then calibrate it conditionally

frankbuckley:
> [!NOTE]

Two days' work on the dispersion priors, in four commits. The full analysis — including the corrections it makes to its own earlier conclusions — is in [`notes/202608020829-kappa-and-eta-q-prior-recalibration.md`](notes/202608020829-kappa-and-eta-q-prior-recalibration.md), structured as an audit trail: later sections correct earlier ones rather than silently editing them, and superseded sections carry pointers.

The commits are separable and worth reading in order. The first changes **how** the dispersion prior is stated; the second changes **what it is a prior about**, and turns up a problem the first could not see; the third is a review response; the fourth finishes the univariate models off.

---

# 1. Two-anchor `kappa` (`d6de1be`, note §18)

The curve is unchanged — `kappa(z) = kappa_min + exp(a_kappa + b_kappa * z)` — but `a_kappa` and `b_kappa` no longer have priors of their own. Priors go on the *age term* at two reference ages in **months**, and the intercept and slope are solved for so the curve passes through both. This is the move the mean trajectory already makes through `slope_anchors`. `a_kappa` and `b_kappa` remain in the trace as deterministics under the same names, so a migrated model's dispersion posterior stays directly comparable with the fits that preceded it, and `kappa_young` / `kappa_old` carry total `kappa` at the anchors.

Recalibrating the old form could not be finished. No setting of its three parameters both admits the slope the data want and keeps young-age dispersion plausible, because the intercept and slope tails compound as `exp(2b)` at the ends of the age range — at `HalfNormal(1.5)` roughly 30% of prior mass puts `kappa` above 200 at `z = -2`. The anchored form also removes the `b_kappa <= 0` sign constraint, states both priors on quantities a per-age empirical fit can check, and makes the prior at any age **exactly invariant** to the pool's age standardisation — the interpolation weight is `(age - young) / (old - young)` in months, so a resample or a study filter cannot move what the prior says.

## Subject-wise subsampling (§§11–13)

`sample_fraction` drew rows rather than children, cutting within-child replication from 1.32 to 1.04 administrations per child and leaving the subject random effect and the observation-level dispersion indistinguishable. VG11 was bimodal at R-hat 1.72 as a result. It now draws whole subjects and keeps all their administrations. The subject key list is **sorted** before sampling: the loader's query carries no `ORDER BY` and `Series.unique` preserves order of first appearance, so an unsorted draw silently depended on what order DuckDB happened to return rows in. Verified — shuffling the input at seed 47 shared only 10 of 30 subjects before the fix.

## VG16 GP anchoring (§16)

VG16 gains the VG10 per-draw GP anchoring at 54 months. It was the one model with subject effects on both `u` and `q` and no anchoring; anchoring takes it from 47 divergences to 0 at `dev` and removes the GP/intercept ridge, leaving `beta_lag` unchanged at 0.308.

---

# 2. Conditional calibration for the random-effect models (`2200d50`, note §19)

**A marginal dispersion calibration and a random-effect model's `kappa` are priors for different quantities.** The marginal estimate answers "how much do counts vary at this age?". A model carrying study and subject random intercepts has already removed most of that variation before its likelihood runs, so its `kappa` describes what is left once a child's own level is known — a much smaller residual. The first commit measured that gap on VG11 at a factor of ten and could not close it; this closes it.

[`scripts/kappa_conditional_calibration.py`](scripts/kappa_conditional_calibration.py) fits the models' own likelihood with the random effects present:

```
logit p_ij = m_c(ij) + s_k(i) + b_i,   b_i ~ N(0, tau^2)
y_ij       ~ BetaBinomial(N_ij, p_ij, kappa(a_ij))
```

Saturated per-age mean (so dispersion is estimated given whatever the mean does), sum-to-zero study fixed effects matching the engines' `ZeroSumNormal` intercepts, subject effect integrated out by Gauss-Hermite quadrature, JAX gradients and L-BFGS. `kappa` is parameterised directly as the floor plus the age term at each anchor, so the Hessian returns standard errors on the quantities the prior states.

Every pool it can measure is **3–10× its marginal counterpart**:

| pool | conditional `kappa` at anchors | marginal | posterior before |
| --- | --- | --- | --- |
| VG11 spoken (12, 20) | **317.5**, **50.5** | 30, 6.7 | 310, 50.0 |
| VG12 understood (12, 20) | **43.0**, **66.4** | 11, 13 | **16.5, 15.6** |
| VG13 understood (12, 17) | **42.2**, **124.1** | 10.8, 13 | **15.9, 14.4** |
| VG13 `q` (12, 17) | **36.0**, **29.7** | 5.8, 4.1 | 40.4, 29.7 |

## The finding is VG12 and VG13, not VG11

VG11 and VG13's `q` were **already right** — their posteriors matched the conditional estimate, so the likelihood was overwhelming a bad prior rather than being distorted by it. Re-centring removes a conflict without moving the answer.

VG12 and VG13's understood outcome were not. Both sat *on* their prior median at ~16 against a data estimate of ~42. That is prior-**dominated** rather than prior-conflicted, so the prior-CDF check that caught VG11 stayed silent — a posterior sitting at its prior median attracts no attention. The cause is structural: both need `kappa` to **rise** with age, which `b_kappa = -b_kappa_mag <= 0` forbids at every parameter setting. The refits prove the constraint was binding:

| outcome | `b_kappa` posterior | 89% ETI | legacy form admitted |
| --- | ---: | --- | --- |
| VG11 | −1.432 | [−1.531, −1.333] | yes |
| VG12 | **+0.211** | [+0.143, +0.279] | **no** |
| VG13 `b_kappa_u` | **+1.397** | [+0.782, +2.083] | **no** |
| VG13 `b_kappa_s` | −0.126 | [−0.277, +0.019] | yes |

Both comprehension outcomes put their entire 89% interval in territory the old parameterisation made unreachable. Not a mis-tuned number — a form that excluded the answer.

## Engine change

Applying this needed the two-anchor form **per outcome** in the joint engines. `build_kappa_for_config`, `kappa_prior_rows`, `kappa_anchor_derived_rows` and `_configure_kappa_priors` now take a suffix that selects both the configuration field and the variable names, with a shared `validate_kappa_fields` rejecting a half- or doubly-specified outcome. `BivariateModelConfiguration` gained `kappa_anchored_u` / `kappa_anchored_s`, and both bivariate build paths route through the shared builder. Outcomes are independent — VG13 anchors both, the DS joint models anchor neither — and a test covers the mixed case no registered model currently uses.

## Validation

`test`-config refits, 4 chains × 2,000 draws. **Every dispersion parameter across all three models now sits at prior CDF 0.41–0.63**, against 1.000 for VG11's two anchors before:

| model | parameter | post mean | prior median | prior CDF | contraction |
| --- | --- | ---: | ---: | ---: | ---: |
| VG11 | `kappa_excess_young` | 311.35 | 311.0 | 0.501 | 0.943 |
| VG11 | `kappa_excess_old` | 44.39 | 44.0 | 0.505 | 0.945 |
| VG12 | `kappa_excess_young` | 38.87 | 40.0 | 0.487 | 0.942 |
| VG12 | `kappa_excess_old` | 62.19 | 63.0 | 0.494 | 0.953 |
| VG13 | `kappa_excess_young_u` | 8.13 | 10.0 | 0.409 | 0.695 |
| VG13 | `kappa_excess_old_u` | 84.63 | 90.0 | 0.473 | 0.894 |
| VG13 | `kappa_excess_young_s` | 31.34 | 33.0 | 0.471 | 0.867 |
| VG13 | `kappa_excess_old_s` | 24.98 | 27.0 | 0.456 | 0.864 |

Posterior `kappa` at the anchors now lands on the conditional estimate: VG11 315.8 / 50.2, VG12 **42.3 / 65.7** (from 16.5 / 15.6), VG13 understood **42.0 / 116.6** (from 15.9 / 14.4), VG13 `q` 35.0 / 28.8.

Three checks had to pass before any of this was read off, and `--recover` / `--mean-sweep` re-run them:

- **`tau` and `kappa` are separable on these designs.** For a child measured once they are confounded, and 84% of VG11's children are measured once. Simulating from a subject-heavy and a dispersion-heavy truth on the real structure returns each correctly, an order of magnitude apart. This **disproves §18's claim** that the two cannot be told apart.
- **The quadrature converges** — but not at the node count a first pass would pick. 24 nodes biased VG11's young anchor down 17%; everything is converged by 160, which is the default.
- **The answer does not depend on the mean model.** Saturated, spline and even a *linear* mean agree within a few percent, so the VG12/VG13 gaps are not an artefact of this estimator tracking the age curve more closely than an HSGP does.

---

# 3. Comprehension: VG02 and VG04 (`6a713ce`, note §20)

The last two univariate models. Neither carries random effects, so the marginal estimator applies and there is nothing to separate — which is why VG02's frame is workable at 346 rows where the DS joint frame of similar size was not. What defeated that one was telling `tau` from `kappa`; with `tau` absent by construction the difficulty does not arise.

| pool | anchors | `kappa` at anchors | stability across mean models |
| --- | --- | --- | --- |
| VG02 understood (DS, n=346) | 18, 36 | **15.4**, **7.1** | 14.8–15.4, 7.1–7.2 |
| VG04 understood (TD, n=1,538) | 12, 18 | **11.8**, **11.3** | 11.6–11.8, 11.1–11.4 |

**Typically-developing comprehension dispersion is flat, and the legacy form could not say so.** VG04's two anchors are near-equal and sit only six months apart — there is no decay to span, so they go where the data are densest (n = 115 and 128) — and the implied slope prior is near-symmetric about zero, `P(kappa rising)` 0.476 against 0.007 for DS spoken. The refit puts `b_kappa` at **−0.022, 89% interval [−0.143, +0.086]** — straddling zero, which `b_kappa = -b_kappa_mag <= 0` makes unrepresentable.

**VG04 against VG12 is the cleanest demonstration that commit 2's distinction is real** rather than an artefact of the estimator. Same outcome, same population, differing only in whether the model has random effects:

| frame | fitted marginally | fitted conditionally | the model's own specification |
| --- | ---: | ---: | --- |
| VG04 (n=1,538) | **11.8** | 42.8 | marginal → prior at 11.8 |
| VG12 (n=5,997) | 11.0 | **43.0** | conditional → prior at 43.0 |

Two frames, two estimators, agreement to within 7% on both readings. The factor of four between the two priors is the specification, not the data.

`Pool` in the calibration script now declares its model's grouping and the estimator mirrors it. That fixed a real defect: `--mean-sweep` had been sweeping a conditional fit regardless of the pool, so a no-random-effects model was being tested for the stability of a fit it never uses.

**Refits** are the best-behaved in the series — 3 divergences each, max R-hat 1.004 and 1.003, min ESS 1,189 and 1,626 — and both posteriors land on the calibration (VG02 14.4 / 6.9, VG04 11.4 / 11.2). VG02's is a live demonstration of the floor/excess ridge: it pulls `kappa_min` to 1.34 and pushes both excesses up while the anchor totals stay within 7%. The decomposition moves; the identified quantity does not.

---

# Review response (`54c0109`)

Three docstring accuracy fixes from the Copilot review: `kappa_min` was still described as an "old-age floor" and the age term as "exponentially decaying", neither of which holds once `b_kappa`'s sign is free. Details in [the reply](https://github.com/dseinternational/vocabulary-growth/pull/192#issuecomment-5159507332); one comment declined with reasoning. Docstrings only, no behaviour change.

---

# What a reviewer should know

**VG09, VG10, VG15 and VG16 are the only models left with uncalibrated dispersion, and that is a limitation rather than a clearance.** On their shared 671-row frame a known `kappa(24) = 317` recovers as 465, 842, 260 and 517 across two outcomes and two seeds — the estimator cannot recover a truth there, so it cannot report one. The presumption from every pool that *could* be measured is that a marginally-calibrated prior on a random-effect model is several times too low, and nothing here rules that out for them. VG15 is excluded for a different reason: its cross-tabulated four-cell frame is not one this estimator reproduces. Stated as such in the note, `PRIORS.md`, the `definitions.py` comment, and asserted by a test.

**The priors are centred on estimates from the same rows the models fit.** Posterior ≈ prior ≈ estimate is *not* independent corroboration of the value; it confirms the prior no longer fights the data and the model can reach what the data say. The evidence the data are placing these posteriors is the contraction (0.94–0.95 on the anchors, against priors spanning roughly [10, 190] at 5–95%) together with the `b_kappa` sign, which no prior here forces.

**None of the three refits passes the convergence gate** (30, 11 and 80 divergences; max R-hat 1.035, 1.017, 1.012). That is unchanged by this work rather than caused by it, and concentrated in `tau` and the GP hyperparameters — every `kappa` parameter has bulk ESS 398–1,357 at R-hat ≤ 1.009. **These fits are not quotable for reporting**; the dispersion estimates within them are what the tables rest on.

**The 16–18 month understood spike is unexplained.** Per-age conditional `kappa` runs 21.6, 23.4, **183.2** at 14, 15, 16 months on 504, 322 and 917 observations — too large and on too much data to be noise, and not a trend. Both understood anchors therefore take `sigma = 0.9` rather than 0.7, so the prior admits a flat or falling curve too. Recorded as the top open item; the obvious candidates (a CDI form boundary, one dataset dominating the 16-month cell) are untested.

**A bug in the calibration's own likelihood was found and fixed.** The normalising constant was out by `2*log(n+1)` per row. It is constant in the parameters, so no estimate or likelihood ratio changes, but the reported nll was not the nll. A test now pins it against `scipy.stats.betabinom`.

**The subject random-effect scales are now the loudest remaining prior-data conflict** — VG13's `tau_subj_q` at prior CDF 0.975, VG11's `tau_subject` pinned at 1.060 ± 0.009 against `HalfNormal(0.5)`. A separate mis-calibration from the one this PR corrects, and untouched by it.

**Several of the note's own earlier conclusions are corrected here.** §17 proposed dropping `kappa_min`; the data reject that at 10–168 log-likelihood units across six pools. The `P(kappa > 200)` criterion the previous calibration was tuned against was wrong — at `n = 810` that is 2.24× the binomial standard deviation, not "near-binomial". §18's "VG11 should not be quoted from this fit" was right for the wrong reason: its dispersion posterior was accurate, and the prior was the defect. And §18's "the data cannot arbitrate" between `tau_subject` and `kappa` is disproved.

**On single-factor comparisons.** VG01 is the only fully clean one — it does not subsample, so the prior is the only thing that changed. VG03's legacy fit predates the row-order sort fix (4,152 rows against 4,075); the shift is under 2% and every change is large against it, but it is not a controlled contrast. For VG12 and VG13 the pre-change posteriors were `dev`-config fits that fail the convergence gate, so the comparison rests on the `b_kappa` boundary behaviour (well mixed: ESS 283 and 372) rather than on the exact `kappa` values.

**Sampling cost is not fully characterised.** Per-draw statistical efficiency is unchanged by the reparameterisation (VG03 dispersion ESS 9,320–10,866 anchored against 10,030–10,199 legacy), but per-draw *cost* is not established because the legacy trace was overwritten. A paired `dev`-config control is an open item.

# Checks

- `pytest`: **415 passed** (66 new across the four commits — the builder's algebra, the configuration contract and per-outcome dispatch, the calibration estimator's recovery in both regimes plus its negative control and node-count bias, the likelihood constant against scipy, the definitions guard, and the subject draw).
- `ruff check src/ scripts/ tests/`, `npm run format:check`, `npm run spellcheck`: clean.
- VG01–VG04, VG11, VG12 and VG13 reports re-rendered; all figure references resolve and no stale `a_kappa_dist` / `b_kappa_mag_dist` images remain.


frankbuckley:
> [!NOTE]

Thanks — three of the four are correct and are fixed in 54c0109. The fourth I think is a false positive; reasoning below.

**`kappa_min` is not always an old-age floor** (`common.py:136`, `definitions.py:111`) — correct, and a good catch. The two-anchor form frees the sign of `b_kappa`, so when dispersion rises with age the exponential term vanishes at the *young* end and `kappa_min` becomes the young-age asymptote. This is not hypothetical: it is why VG13's `kappa_min` is 30 rather than 3, since a third of its 8–18 month frame sits below the young anchor where the asymptote alone sets the level. The behaviour was already written up in the note and `PRIORS.md`; the docstrings simply had not been updated to match, which is exactly the kind of drift worth flagging. Reworded in `AnchoredKappaPriors.kappa_min_dist`, `KappaAnchorPriorParams.kappa_min_mu`, the two `excess_*` fields ("above the floor" → "above the asymptote"), and `build_kappa_of_z_anchored`.

**"exponentially decaying age term"** (`definitions.py:79`) — same issue, same fix. Now "an asymptote plus an exponential age term". Worth noting the decay-only reading is precisely what the legacy `b_kappa_mag >= 0` enforced and what this PR removes, so leaving the word in the class that removes it would have been actively misleading.

**Stray leading quote** (`common.py:240`) — the `""""` was intentional (the docstring opens with the quoted table name "Derived quantities"), but it reads as a typo and is easy to misparse. Rewritten as `Rows for the "Derived quantities" table, ...`.

**Docstring indentation** (`tests/test_gp_utils.py:10`) — declining this one. It is a module-level docstring, and every line of it is at column 0, which is both internally consistent and what PEP 257 expects there; the continuation lines have nothing to align to. I think this comment is applying function-docstring conventions to a module docstring. Happy to revisit if I have misread it.

No behaviour changes in any of this — docstrings only. 415 tests still pass.

PR #195, opened 2026-08-03
feat(data): consume the es_01 Galeote Spanish dataset

frankbuckley:
> [!NOTE]

Wires the Spanish CDI-Down dataset (`es_01`, Galeote) into the data pipeline. Adds **186 Down syndrome observations** at ages 11–71 months, taking the analysis view from 1,221 to **1,407 rows**, and **185 signed observations**, making it the second-largest signing source in the pool. Prepared upstream in [dsegroup/research-data-analysis#10](https://github.com/dsegroup/research-data-analysis/pull/10) and [#11](https://github.com/dsegroup/research-data-analysis/pull/11).

The source is cross-sectional (one observation per child) and is the sample reported in Galeote, M., Sebastián, E., Checa, E., Rey, R., & Soto, P. (2011). The development of vocabulary in Spanish children with Down syndrome: comprehension, production, and gestures. _Journal of Intellectual & Developmental Disability_, 36(3), 184–196. <https://doi.org/10.3109/13668250.2011.599317>

## Wiring

`scripts/prepare_data.py`: source entry, merge block (`study = 12`), DuckDB `vocab_es_01` table. `src/vocab_growth/data_utils.py`: a `vocab_combined` UNION, plus the CDI-Down entry in the native-ceilings note.

## Symbolic gestures are the signed lexicon

The CDI-Down "added a third column to assess the comprehension and production of gestures representing specific lexical items" (Galeote et al., 2011), and the study set out to analyse vocabulary size "both in the oral and gestural (symbolic gestures) modalities". These are **symbolic or referential** gestures, each tied to one of the 651 words — a gestural lexicon scored per word, which is the construct VG14/VG15/VG17 estimate. So `signed` is es_01's gestural total and `produced` its recorded spoken-or-gestured union.

Both fit the existing conventions without re-derivation:

- the gestural count is a **total**, covering words gestured whether or not they are also spoken — matching `uk_02` and `nz_01` rather than `uk_01`'s signed-only field (`SIGNED_ONLY_STUDIES`);
- `produced` is a **de-duplicated union recorded by the source outright**, where `uk_02`'s and `nz_01`'s must be built from mutually exclusive cells. It exceeds `spoken` by a mean of 28 words.

`model_vg18`'s `produced` taxonomy and its recommended union-study subset are updated to include es_01, as is `model_vg17`'s example.

One caveat to note when interpreting the signing models: symbolic gestures are not a formal sign language, so es_01's signing construct is a near neighbour of the taught-sign lexicons in `uk_01`, `uk_02` and `nz_01` rather than the identical thing.

### Guard on one impossible row

A gestural total larger than the union it belongs to is impossible — a union cannot be smaller than either of its parts — so such a row's `signed` is masked rather than passed to the signing models as a total. This is expressed as a **rule** in the view rather than keyed to a subject id, so it survives a corrected or extended source. One row of 186 matches (1 word spoken, 15 gestured, union 11); which of its three source numbers is wrong cannot be determined, its `understood`/`spoken`/`produced` are unaffected, and the row is retained.

## The typically developing group is excluded from `vocab_combined`

The CSV also holds 186 TD children matched pair-wise to the DS children on mental age and sex (`pair_id` links a pair; mental age level identical for all 186 pairs, Brunet-Lézine total within 0.3 months for all 186, sex identical for 185). The view admits `group = 'DS'` only. They are a Spanish-normed sample on a different instrument from the Wordbank forms `load_data` draws on, so pooling them would put a second instrument into the reference pool the Down syndrome exclusions are benchmarked against — `TD_POOL_EXCLUDED_DATASETS` exists to keep that pool clean. They stay in the `vocab_es_01` table, so a matched-pair analysis is available whenever it is wanted; that is a bigger question than wiring a source and wants its own change.

## Ceiling

`survey_vocab_max = 651` — the CDI-Down vocabulary checklist is 651 words in 21 categories. Two children sit **exactly at** the comprehension ceiling; these are legitimate but censored (their true receptive vocabulary is at least 651), and the form-ceiling guard keeps them, since it drops only counts strictly above the ceiling. `scripts/data_coverage.py` puts es_01's near-ceiling comprehension exposure at **7.0%** (13 of 186) and spoken at 2.2% — below the small-ceiling sources `methods-data.qmd` flags at about a fifth, so this does not worsen the harmonisation exposure.

## Not wired: the four-cell cross-tab

es_01 also supports a **four-cell within-understood cross-tab** — neither, spoken-only, gestured-only, both — of the kind `common_joint_modality` builds for `uk_02`, since `both = spoken + gestured − union` and `neither = understood − union` are both derivable. That would let es_01 inform the sign–speech overlap `psi` directly instead of only through its marginals. Left for a follow-up: it is a model change rather than a data change, and the one inconsistent row yields a negative cell.

## ⚠️ Every model of record is now stale

Adding a CSV to `data/` changes `source_data_hash`, so `sync_report_figures.py` will refuse to sync until the models are refitted, and `scripts/vg07_study_effects.py` will trip its label-count assertion against an old trace. **A full DS refit and a findings note are the required follow-up**, as for the `ie_02` integration (`notes/202606270930-ie02-refit-and-findings.md`, #77). The signing models change most here — es_01 is a large new signing source.

I deliberately did **not** touch `docs/report/`. The study inventory table there is generated from the data, so it picks es_01 up on the next render, but the Measures prose describes the pool the _reported fits_ used — amending it before the refits would describe a pool no published fit reflects.

## Verification

- `pytest` — **439 passed**, including `test_every_registered_model_builds` for every registered model against the rebuilt es_01-inclusive database.
- Four regression tests pin the DS-only filter, symbolic gestures mapping to `signed` with the recorded union as `produced` (not `spoken + gestured`), the impossible-total guard, and the 651-item ceiling (at-ceiling kept, above-ceiling dropped).
- `ruff check src/ scripts/` clean; `npm run format:check` clean; `npm run spellcheck` 0 issues (five proper names added to the allow-list).
- `scripts/prepare_data.py` and `scripts/data_coverage.py` both run end to end; es_01 contributes 186 of 372 source rows to the view, with `signed` present on 185 and no row where `signed > produced`.


frankbuckley:
> [!NOTE]

Remapped in 2e7a393, and the PR description is rewritten to match — my original wiring had this the wrong way round.

The maintainer supplied the paper's own account: the CDI-Down "added a third column to assess the comprehension and production of gestures representing specific lexical items", and the study set out to analyse vocabulary size "both in the oral and gestural (symbolic gestures) modalities". These are symbolic/referential gestures tied to individual words — a gestural lexicon, not pointing and waving — so they are the `signed` construct, and my earlier reasoning (that gestures for a word were not a signed lexicon, so `signed` should be NULL and `produced := spoken`) was wrong.

`signed` is now the gestural total and `produced` the recorded spoken-or-gestured union. Both fit the existing conventions without re-derivation: the gestural count is a **total** including words also spoken, so it belongs with `uk_02`/`nz_01` rather than `uk_01`'s signed-only field, and the union is recorded by the source outright where `uk_02`'s and `nz_01`'s have to be reconstructed from mutually exclusive cells.

What this changes:

- es_01 becomes the **second-largest signing source** at 185 usable signed observations — all 186 rows carry a non-zero gestural total, mean 41, max 214, signed ratio `signed/understood` mean 0.185.
- `produced` gains a mean **28 words** over `spoken` (max 119), so es_01 joins `uk_01`/`uk_02`/`nz_01` in `model_vg18`'s de-duplicated-union group. That docstring and its recommended study subset are updated, as is `model_vg17`'s example.
- The one impossible row now matters, where before only its `spoken` reached the pool. A gestural total above its own union cannot be a total, so the view masks that row's `signed` — as a rule, not keyed to a subject id, so it survives a corrected source. It hits 1 row of 186 and leaves `understood`/`spoken`/`produced` intact.

Two things left deliberately undone. Symbolic gestures are still not a formal sign language, so es_01's signing construct is a near neighbour of the taught-sign lexicons rather than the identical thing — documented, not masked. And es_01 turns out to support a full **four-cell within-understood cross-tab** (neither / spoken-only / gestured-only / both), the structure `common_joint_modality` already builds for `uk_02`, which would let it inform `psi` directly rather than only through its marginals. That is a model change rather than a data change, and the one inconsistent row yields a negative cell, so it wants its own PR.

439 tests pass; `ruff`, `format:check` and `spellcheck` clean.

PR #209, opened 2026-08-09
Trace persistence tiers: cut artifact size without touching inference

frankbuckley:
> [!NOTE]

## Why

A full reporting output tree is ~257 GB and essentially all of it is `trace.nc`: the figures, tables, diagnostics and rendered HTML together come to ~0.65 GB. Individual rep traces run to VG11 53.6 GB, VG13 40.6 GB, VG12 21.3 GB. This is not only inconvenient — VG12's recovery replicate r08 died with `ENOSPC` on the VM's `/scratch` on 2026-08-06 partway through writing its own trace.

Measured on two completed fits, the free parameters come to **~0.1 GB**; everything else is a deterministic function of them evaluated at each observation, grid point or subject. Dropping the recomputable half costs nothing statistically. The alternative lever — subsampling the typically-developing pool — was considered and rejected in `202608050900-td-hierarchical-geometry.md` §10, because within-child replication is the binding constraint on the TD hierarchical geometry and any subsampling removes repeat-measured children.

Design rationale is in `notes/202608081445-trace-persistence-tiers.md`, which is on #208 rather than this branch (the two are independent; this one is branched from `main`).

## What changed

**A shared save.** Seven call sites across six modules each wrote the trace themselves, so any policy would have had to be repeated seven times and would have drifted. They now route through one `save_trace`; there are no raw `to_netcdf` calls left under `models/`.

**Three tiers.** `full` (default), `compact` (drop observation-sized posterior deterministics and the duplicated scaled random effects), `minimal` (also drop the stored `log_likelihood` and `posterior_predictive`). Selection is by *dimension*, not a name list, so it survives new models: `obs_id` (and joint models' `obs_*_id`) and the concatenated `all_id` grid go; `plot_id`, `query_id`, `study_id`, HSGP coefficients and free scalars stay.

**`--trace-persistence` on `fit_model.py`**, resolved with the same precedence as `--output-dir`: flag > `$DSE_VOCAB_GROWTH_TRACE_PERSISTENCE` > `full`. Every other fitting entry point honours the environment variable without needing its own flag — which matters for `fit_recovery.py`, since a recovery replicate is what hit `ENOSPC`.

**The manifest records what was written**, under `artefacts.trace`, so a reader finding no `f_obs` can tell "dropped by policy" from "truncated or corrupt".

**`loso_compare.py` refuses a compacted trace up front**, before reading it.

## Measured

| | VG10 trace | |
| --- | --- | --- |
| `full` | 9.82 GB | |
| `compact` | **3.16 GB** | −68% |
| `minimal` | **1.95 GB** | −80% |

The compacted trace produces **bit-identical** reporting output: `ages`, `p_s`, `kappa_s`, `p_u` and `n_trials` all compare equal against the full trace through `comparison.load_outcome_trajectory`.

## For the reviewer

- **The sharpest edge is that filtering is group-scoped, not global.** `observed_data/y_obs` and `constant_data/X_obs` are *also* `obs_id`-dimensioned — an unscoped dimension rule would have deleted the data itself. Worth checking that reasoning specifically.
- **An inapplicable tier raises rather than silently writing full.** A test I wrote for a different purpose showed that a `compact` request on an unreadable object planned nothing and wrote the full trace, producing an artifact that looks correct while the policy had done nothing.
- **Two saves are pinned to `full` on purpose**: the convergence-failure preservation save, because that trace exists only to be investigated and a post-mortem should not have to work around a storage policy.
- **The setting is deliberately not on `ModelDefinition`.** Definition fields are part of the model graph and its fingerprint — `td_languages` is one precisely because changing it forces a refit — whereas this changes nothing about the posterior, so putting it there would invalidate every fitted model for a storage decision.
- **`kfold_loso.py` deliberately has no guard.** It reads `p_u_obs` / `p_s_obs` / `q_obs`, but from the in-memory trace returned by its own `fit_fold` — it fits each fold itself and never opens `trace.nc`. My initial audit had it wrong; `loso_compare.py` was the only real disk consumer.
- **Backward compatibility:** fits written before this carry no persistence record and are treated as `full`, so nothing already on disk is refused for lacking a field.

## Verification

566 tests pass (29 new); ruff, Prettier and CSpell clean. Verified end to end through the real pipeline, not only in unit tests: a dev fit of VG01 at `compact` drops 7 variables, records them in the manifest, keeps `observed_data`/`constant_data` and the plot grid, and still passes `validate_fit_output`; the same fit at the default writes 124.39 MB against the compacted 40.73 MB and records `full`. Both were redirected to a scratch output root, so no model of record was touched.

## Not in this PR

The §7 rollout — fitting one model at each tier and confirming the reporting path is byte-identical before flipping the default. **The default stays `full`, so nothing changes for existing work until someone opts in.** A test asserts `save_trace(..., "full")` is identical to the previous call.


frankbuckley:
> [!NOTE]

Thanks — all seven comments were valid and are addressed in 75a6b1d. I verified each against the code rather than taking them on trust; four were substantive.

**The `z_`-named effects (`common_joint_modality.py`) — the most significant.** Correct, and worse than it first looks: the de-duplication never fired on the joint-modality or trivariate engines at all, because they name the offset for its distribution (`delta_u` from `z_u`, `delta_subj_*` from `z_subj_*`) rather than `*_raw`. So both copies survived compaction on **the very model whose 1.20 GB of duplicated subject arrays I cited as the motivating evidence**. The rule now recognises both conventions.

It also needs a dimension match, which the comment didn't mention but which matters: VG15's `delta_sign` is `z_sign` — indexed over sign-informed studies only — scattered into a zero-filled vector over *all* studies. That is not an elementwise scaling and cannot be rebuilt from `z_sign` and a scale, so requiring identical dims correctly keeps it. Confirmed against the real trace: `delta_sign` has dims `(study_id,)` against `z_sign`'s `(z_sign_dim_0,)`.

**`regenerate_plots.py`.** Correct, and a genuine gap in my audit. I checked that `extract_model_samples` runs in-process on the in-memory trace during a fit, and missed that `regenerate_plots.py:157` loads the saved trace from disk and calls it. Now guarded before the load.

**Parameter-recovery scoring.** Correct, and the failure mode is the bad one: `target_variables` intersects truth with posterior, and `_raw` offsets are deliberately excluded from scoring while the scaled effects *are* the targets — so a compacted recovery fit would drop `delta*` from the target set and report the remainder as if it were the whole. Silent degradation, not a crash. Guarded in `recovery/compare.py`.

**Startup validation.** Correct — the comment claimed something the code did not do. With the flag omitted, `set_trace_persistence(None)` only clears the override, so an invalid `$DSE_VOCAB_GROWTH_TRACE_PERSISTENCE` went unparsed until the first save, hours into a reporting fit. Now resolved at startup:

```
$ DSE_VOCAB_GROWTH_TRACE_PERSISTENCE=smallish python scripts/fit_model.py vg01 --config dev
ValueError: $DSE_VOCAB_GROWTH_TRACE_PERSISTENCE is 'smallish'; expected one of full, compact, minimal.
```

**The three broken note links.** Correct at the time. #208 has since merged, so `notes/202608081445-trace-persistence-tiers.md` is on `main`; I've merged `main` into this branch and the links now resolve.

574 tests pass (8 new, covering both naming conventions and the scattered-effect exception). Ruff, Prettier and CSpell clean.

Issue #213, opened 2026-08-12
Full refit and report publication after uk_07 integration (needs a larger VM)


frankbuckley:
> [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Opus 5).

Coordination issue for the full refit and report publication that #212 makes necessary. It cannot be completed on a laptop: **VG11 peaked at 157 GB anon-RSS** in the August run and will not fit in 48 GB. This sets out the state, the sequence, and the gotchas, so the VM run is a checklist rather than a rediscovery.

## Why everything is stale

#212 adds `data/vocab_data_uk_07.csv`. `source_data_hash` fingerprints **every** CSV in `data/`, so all fifteen models of record are invalidated — including the five typically-developing ones, whose analysis frame uk_07 does not touch. The hash cannot express "this source is Down-syndrome only", and that is the correct conservative behaviour rather than a defect.

VG15 is stale twice over: its model graph also changed (`delta_psi`, `tau_psi`, `include_es01_cells`), which `fit_artifacts` detects independently as "The model definition differs from the current registered definition."

## Current state

Eight DS models were already refitted locally from a clean tree at `rep` and are publish-valid, but they were fitted **before** #212's VG15 changes, so the DS pool must be re-run as a set on the VM for a consistent provenance stamp.

| model | state | note |
| --- | --- | --- |
| VG01, VG02, VG05, VG07, VG10, VG14, VG16 | refit locally, valid | re-run on VM for one consistent batch |
| VG09 | refit locally, valid | needed hightune (tune 12000 / draws 8000 / `target_accept` 0.97) |
| VG08 | refit locally | needed **stronger** hightune (tune 16000 / draws 10000 / `target_accept` 0.99); plain `rep` and the August hightune settings both failed the gate |
| VG15 | **must refit** | graph changed in #212 |
| VG03, VG04 | stale | data hash only |
| VG11, VG12, VG13 | stale | data hash + pre-existing soft-tier caveats; publish via `--allow-caveats` |

## Measured timings (48 GB / 12 performance cores, sequential)

VG02 7m · VG01 9m · VG10 16m · VG16 16m · VG07 17m · VG05 17m · VG15 22m · VG14 24m · VG09 hightune 33m · VG08 strong hightune 140m. **DS pool ≈ 5 h sequential.** August's record puts VG11 at 3h29m alone.

## Sequence

[...]

Issue #190, opened 2026-07-31
State of play: item-difficulty workstream complete; refit, owner decisions and report corrections outstanding


frankbuckley:
> [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Fable 5).

> [!IMPORTANT]
> **Closed 2026-09-07.** Nothing left here needs a fit. The owner decisions (A, C.2, the proper-noun denominator opened by C.1) and the caveats table gated on them (B, last item) are now tracked in #320; the refit is #289; #147 and #185 are closed. See the closing comment.

Coordination issue for the item-difficulty and Edgin-audit workstream completed 26–27 July 2026. It records what landed, the position it leaves the project in, and the work still outstanding with its dependencies. Individual work items are tracked separately where they already have issues (#281 — which supersedes the since-closed #186 — #185, #147); the untracked items are listed here with checkboxes so they can be split off or actioned directly.

## What landed

| PR | What it did |
| --- | --- |
| #182 | Corrected an inverted headline claim, surfaced the soft convergence caveats, repaired the two ie_01 data defects |
| #183 | Re-billed the item-difficulty work on Rasch sufficiency; audited the `us_01` (Edgin) subset and registered the inverse sensitivity; pre-registered the Route 1 DIF analysis; moved the DuckDB build ahead of pytest in CI |
| #184 | Removed the orphan `us_01` CSV and guarded the source registry |
| #187 | Whole-month expected-count tables and figures for every model |
| #188 | Renamed the misleading `joint_trajectory_hdi` figure |
| #189 | Dependabot: `setup-micromamba` 3.0.0 → 3.1.0 |

`main` is green on every one. The PR queue is empty.

### The item-difficulty question is closed, and closed in an unusual way

The work began as "item exchangeability is false, and therefore the likelihood is questionable". The study owner challenged that framing with the Rasch sufficiency argument, and it held. Item exchangeability **is** false — a model-free hypergeometric test against the three DSE checklists rejects it decisively — but it is **nearly costless for the quantity the models fit**: under a Rasch-type item model the total score is a sufficient statistic for ability, so heterogeneous difficulty can reach a sum-score model only through the distribution of the total, and that channel carries 0.77%–5.27% of total variance in VG10, where even a 40% error moves the total standard deviation by about 1%.

[...]

PR #214, opened 2026-08-12
feat(sensitivity): register the three checks the refit would otherwise miss

frankbuckley:
> [!NOTE]

Three questions a reader will ask of the next published run had no registered variant, so nothing would have run them. Each would otherwise have needed its own VM window after the fact. Registering them costs nothing now and adds six sensitivity fits to the run planned in #213.

## 1. `dse-native-only` (VG10, VG15) — the 810-item reference denominator

Closes the item #190 calls "the highest-value untested model check outstanding".

Every model scores raw counts against `n_trials = 810`, so a 416-item Oxford CDI count enters on a denominator its form never used. That is sound only if the shorter forms hold the easier items — and the sufficiency result in `notes/202607261540` is the proof that no aggregate analysis of these data can test it, since a statistic sufficient for ability carries no information about item composition. So the assumption is probed by deleting the rows that need it, not by modelling it differently.

278 of 1,521 rows survive, from 194 children across ie_01 (its 810 wave only), ie_02, uk_02 (DSE arm only) and uk_06, spanning 9–115 months. Comprehension is least affected — 259 of 987 understood observations survive against 264 of 1,428 spoken — because the short forms are production-heavy.

This one needed a real data path rather than a registry line. Three of the four cross-tab sources read their own CSVs, so a filter on the merged view would never have seen them and uk_07, es_01 and nz_01 would have re-entered through the side door carrying exactly the harmonisation the variant exists to remove. Both engines now gate those blocks, and both print `Non-native-ceiling rows excluded` (1,243 against the current pool) so a variant that stops biting is visible rather than reassuring.

On VG15 the only surviving cross-tab is uk_02's DSE arm — all 56 of its four-cell rows are that form — so `psi` falls back to its single-study branch with no `delta_psi` or `tau_psi`. That is documented on the flag and in the runbook: the variant answers the denominator question and the between-study question together and cannot separate them, so it is a trajectory-shape check, not a `psi` check.

## 2. `tau-psi-narrow` / `tau-psi-wide` (VG15)

`tau_psi_sigma = 1.0` was set from the measured between-study spread, which makes it data-informed rather than externally justified — precisely the condition that created Target 8 for every trajectory anchor. With four informed studies `tau_psi` is weakly identified, so the prior does real work on how far the per-study values shrink toward the reported centre, and therefore on the headline `psi`. Narrow (0.3) forces near-pooling, roughly the pre-2026-08-12 behaviour; wide (2.0) lets each source sit where its own cells put it.

## 3. `psi-drop-es01` / `psi-drop-uk07` (VG15)

Both inclusion flags already document that setting them `False` isolates that source's pull on `psi`. Nothing ran it. Each variant drops one source's cross-tab while keeping its marginals, so U, q and r are unchanged and only the association loses evidence — the contrast attributes movement to that source rather than to "fewer cells in general". es_01 is the one that matters most: 185 of the 434 `psi`-informing rows, and the only source sitting at independence while the other three are well above it.

## Verification

- Both engines build under every new variant; the VG15 native fit correctly takes the single-study `psi` branch (no `tau_psi`, no `z_psi`, `psi_study` still reported).
- Prepared-frame check on the real sources: base 1,431 rows / 323 four-cell from three studies; native 265 rows / 56 from uk_02 only; `psi-drop-es01` 323 → 138 with the row count unchanged; `psi-drop-uk07` 323 → 241.
- New tests cover the filter in isolation, the real pool's surviving counts, the cross-tab gating through the real engine, and that each variant differs from its base.
- Full suite green; ruff, Prettier and cspell clean.

No default changes, so no existing fit is invalidated by this PR.

Refs #190, #213.


frankbuckley:
> [!NOTE]

**Correction to this PR's description.** It says "No default changes, so no existing fit is invalidated by this PR." That is wrong, and the mechanism is worth recording.

`validate_fit` compares the manifest's recorded `model.definition` payload against `normalise_for_json(<current definition>)`. That payload is the **whole field set**, not a description of the model graph — so adding a field re-stales every model on that dataclass even when its default changes nothing about what is sampled:

```
$ python - <<'PY'
from vocab_growth.fit_artifacts import normalise_for_json
from vocab_growth.models.definitions import VG10
import json
cur = normalise_for_json(VG10)
old = json.load(open("output/models/VG10-.../fit_manifest.json"))["model"]["definition"]
print("only in current:", sorted(set(cur) - set(old)))
print("changed values:", {k: (old[k], cur[k]) for k in set(cur) & set(old) if old[k] != cur[k]})
PY
only in current: ['dse_native_only']
changed values: {}
```

So `dse_native_only` landing on `BivariateModelDefinition` and `JointModelDefinition` moved VG05, VG07, VG08, VG09, VG10, VG13, VG15 and VG16 to `The model definition differs from the current registered definition`. VG01 and VG02 are unaffected only because they are univariate.

**Nothing is lost here.** Every one of those models was already invalid on the `source_data_hash` change from #212, and the run tracked in #215 refits all fifteen as one batch for a consistent provenance stamp. The cost of this PR remains the eight sensitivity fits it registers.

The general point is the part worth keeping: **the definition fingerprint is field-set sensitive, not graph sensitive.** An inert flag added to a shared definition class is not free — it invalidates every model on that class. That is the conservative and correct behaviour for a provenance check (it cannot be expected to know which fields are load-bearing), but it means such a field should land with a refit already planned, as this one did, rather than being assumed harmless.

Issue #242, opened 2026-08-23
Correct and revalidate the VG16 cross-lag model


frankbuckley:
> [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

## Summary

A code, data and statistical review of VG16 found a definite row-order-dependent lag bug and several model-output and interpretation problems. No estimate should be presented as the current VG16 result until the lag construction is corrected, the estimand is clarified and VG16 is refitted; historical and diagnostic estimates may still be quoted with their provenance and caveats.

The review is documented in `notes/202608231714-vg16-statistical-model-review.md` in [PR #243](https://github.com/dseinternational/vocabulary-growth/pull/243).

## Confirmed defects

- `_compute_prev_wave_lag()` advances its source state row by row rather than by complete child-age groups. When a later recorded age has different-form rows, the result depends on arbitrary tie order. On the current frame, 66 spoken observations from 46 children receive no lag even though they would receive one under a grouped retain-all interpretation: the implementation gives 409 spoken rows a lag, against 475 under that interpretation. A form-selection or collapse policy could produce a different valid total, but the current order-dependent mixture implements neither policy consistently.
- The same lag algorithm is copied into the two VG16 experiment scripts, so the supporting analyses reproduce the defect.
- Subject-marginal spoken predictions use `q_plot`/`q_query` without `beta_lag` or a simulated lag history. They are valid predictions conditional on no usable lag history, but the report presents them as longitudinal/unseen-child marginal predictions without disclosing that conditioning.
- Pointwise understood PSIS-LOO leaves the supposedly held-out understood count embedded in later spoken lag predictors. The reported score is therefore not clean leave-one-understood-out prediction.
- The generic pair plot omits `beta_lag`, despite the report saying that the plot shows its trade-offs with the random-effect scales.

## Statistical and reporting problems

- The population-relative predictor mixes persistent between-child standing with temporary within-child deviation. With independent understood and production-ratio child effects, `beta_lag` can proxy omitted persistent covariance. VG16 can support the sign of its model-conditioned prospective association, but it cannot by itself establish a distinct within-child dynamic, causal mechanism or causal direction.
- The report's claim that the fitted coefficient is a lower bound relies on a classical errors-in-variables correction whose assumptions are not established for a clipped logit Beta-Binomial predictor in this nonlinear multilevel model.
[...]

Issue #238, opened 2026-08-23
Address VG14 and VG15 statistical-review findings


frankbuckley:
> [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

## Summary

A read-only review of the current VG14 and VG15 implementations found several confirmed reporting and specification defects, alongside statistical risks that require sensitivity analysis before the VG15 association estimates are treated as publication-ready. VG14 remains appropriately superseded by VG15, but its retained implementation should either be corrected or labelled more explicitly as non-substantive.

## Confirmed defects

- Apply the 72-month understood-vocabulary evidence cap consistently to q, r, p_any, crossover outputs, tests, downstream comparisons, and report lookups.
- Stop reporting the movable tent knot as the actual peak of the full signed-ratio curve. Either rename the estimand or calculate the per-draw peak of the complete curve with boundary diagnostics.
- Centre VG15 understood study effects over understood-informed studies only; nz_01 currently enters the zero-sum constraint despite having no understood likelihood contribution.
- Correct VG15 milestone logic so it detects genuine transitions and treats grid-boundary maxima as censored.
- Recalibrate VG14 dispersion priors if VG14 is retained for inference. The current priors were calibrated for a random-effects graph that VG14 does not contain.
- Correct stale descriptions of the fitted models and validation coverage.

## Statistical validation work

- Quantify the effect of repeated-child dependence in the VG15 composition likelihood using a child-shifted cell model, one-administration-per-child analysis, cluster bootstrap, or equivalent sensitivity.
- Add quantitative validation for cells_obs and nz_prod_cells_obs, including per-source and age-banded posterior predictive checks.
- Add honest child-level held-out assessment for VG15 rather than relying only on marginal pointwise LOO.
- Run multiple current-graph reporting-quality recovery replicates, with particular attention to population and study-specific psi shrinkage.
- Assess whether the four-cell and three-cell likelihoods require separate concentration parameters.

## Acceptance criteria

- Corrected code, tests, reports, and downstream scripts agree on estimands and reporting ages.
- Derived peak and milestone summaries distinguish observed transitions from censored or initially true states.
- A current reporting-quality VG15 fit passes convergence checks.
- Association-likelihood calibration and repeated-child sensitivities are documented with interpretable results.
- Publication-facing text states the supported estimand and remaining limitations without claiming validation that has not been run.

PR #249, opened 2026-08-24
Remove the fit-memory ceiling, register VG21/VG22, and contrast individual growth with the median

frankbuckley:
> [!NOTE]

Twenty-two commits spanning three strands: a memory lever that removes the ceiling on reporting-quality fits, the VG21/VG22 model registrations with VG22's identification fix and rank decision, and two new figures that contrast individual growth with the population median. Everything statistical is either provably inert or recorded with its evidence.

> [!IMPORTANT]
> This branch was cut before the 19 statistical-review commits that landed on `main` on 2026-08-23, and **it conflicts with them in five files**: `docs/models/README.md`, `notes/README.md`, `src/vocab_growth/models/common_bivariate.py`, `src/vocab_growth/models/common_bivariate_re.py`, `tests/test_gp_utils.py`. Every one is an additive collision — both sides appended index rows, tests, or engine changes to the same regions — but they need a merge before this is landable. Upstream's engine commit is explicitly fit-preserving (the HSGP centre was verified equal to the previously computed one), so the merge should not invalidate anything.

## Memory: stop sampling the observation-sized deterministics

Fit memory was dominated by observation-sized deterministics stored per draw, scaling as `n_obs × draws` — VG11's hightune was killed at 247 GB and VG13's `rep` fit plateaued near 178 GB. The engines now pass `var_names` to `pm.sample`, so nutpie never evaluates them. The model graph is untouched, so the draws are identical: on VG10 at `dev`, 48 of 60 shared posterior variables are bit-identical and the other 12 agree to 1e-11, with the stored `log_likelihood` bit-identical to the last element.

Measured on VG10 at `test`: peak RSS 5.45 GB → 3.33 GB, `trace.nc` 2,450 MB → 879 MB, posterior group 2,072 MB → 500 MB. **This is a memory result, not a speed one** — wall time moved about 4%. No `rep`-scale fit has yet been made on the new code, so the extrapolation to the TD pools is exactly that.

One consequence had to be decided rather than absorbed: ArviZ derives PSIS-LOO's `reff` from the mean ESS over *every* variable in the posterior group, so it moved when the storage policy moved (VG10: 0.5613 before, 0.5540 after, 0.9047 over the sampled parameters alone). `reff` is now pinned to the model's free random variables, recorded in the trace, and every `az.loo`/`az.compare` in the project goes through it. A trace that predates the record falls back to ArviZ's convention **and says so out loud**.

## Models: VG21, VG22, and the rank decision

VG21 registers the adopted window-22 specification; VG22 adds a low-rank factor over all four child effects, with `rank` as a definition field because the residual likelihood cannot choose it (rank 2 is within 2.60 on 2 df of rank 3, and rank 3 within 0.0000 of the free 4×4).

VG22's first `rep` run failed its convergence gate on two of three ranks, and the cause is worth reading: the rotational gauge was anchored on `b1u`, the one child effect with no between-child variance. With `tau ≈ 0` that row of the loading matrix is ≈ 0 whatever its direction, so the sign constraint pinned nothing and the chains split into exact mirror modes — 723 parameters over R-hat at rank 2, and a continuous rotation ridge at rank 3. The fix permutes the anchor order onto effects with demonstrated variance and is a pure gauge change: `Σ`, its positive semi-definiteness, `Σ_ii = tau_i²` and the free-parameter counts are all preserved, and the rank-1 graph is identical. **The general lesson is recorded in the note: an identification constraint is only as strong as the parameter it sits on.**

The registered default then moved from rank 2 to rank 3, on the refits rather than the gate: rank 3 cleared the gate on plain `rep` with 8 divergences where rank 2 needed a hightune to reach 18 (178 without one), and the two disagree on the spoken child-slope scale by about 2.8 combined standard errors on shared data. **Not a promotion of VG22 to model of record** — VG20 holds that role and is untouched. Both stored VG22 fits are now stale and neither can be relabelled into its new role, so a canonical rank-3 fit and a rank-2 variant fit are outstanding.

## Marginalising the singleton child effects: exact, and rejected

Integrating out the child effects that only one row ever sees is exact, and the implementation is verified so: equivalence with the explicit arm holds on every shared scalar to |z| ≤ 0.85 and on the published trajectory to 0.00126 in probability. It shrinks the sampled space from 5,850 dimensions to 1,031 and lifts minimum energy BFMI from 0.187 to 0.492 — confirming, with the posterior provably unchanged, that the singleton children really are a substantial part of what suppresses energy exploration in the TD hierarchical models.

**It is not adopted.** Break-even needs a marginalised gradient to cost at most 1.80× an explicit one; it costs 10.5× measured in situ and 22× by microbenchmark, so the explicit arm delivers between 5.9× and 12.2× more effective samples per unit work. The code, the tests and the note stay; `singleton_marginalisation` is `None` on every registered definition, so no fitted output is affected. One pre-registered prediction was falsified and is recorded as such — divergences rose 20 → 48 rather than falling — and doubling the quadrature nodes leaves the answer alone while making the sampling worse, so the node count is a tuning parameter with a cost on both sides rather than a free accuracy dial.

## Figures: individual growth against the median

Two overlays on the median-trend plots. The observed one joins each child's own observations for children with three or more administrations. Its hazard is the recording form: the Down syndrome pool spans item counts from 396 to 810 and roughly half the multi-visit children change form, so a child near the ceiling of a short form can record *fewer* words on a longer one a month later — one child here scores 393 understood on a 416-item form at 47 months and 347 on an 810-item form at 48. Segments are therefore solid within one form and **dashed across a change**, near-ceiling observations are marked, and the legend states the figure's own composition.

The predictive one draws the trajectories the bands only summarise. The draws were already coherent — `sample_posterior_predictive` reuses one child effect across the grid on purpose, with the right construction for each model — and that structure was being discarded at summary time. What is drawn is the *expected* unseen-child curve rather than `y_*_plot`, because the stored counts carry noise drawn independently at each grid point that moves ~90 words between adjacent points against ~1.3 for the curve beneath.

Together they show something worth looking at: VG20's predicted children preserve rank order almost perfectly (0.9916 between 20 and 60 months), which is what an intercept-only child effect implies, while the observed children visibly differ in steepness. That is the clearest argument yet for the child-slope structure VG19 and VG22 carry.

## Three defects fixed that misreported rather than failed

- `aggregate_summary` died on every run with `TypeError: 'tuple' object is not callable` — `groups` is a property on ArviZ 1.x's `DataTree`. **Dropping the parentheses would have been worse than the crash**: the entries are paths, so the membership test finds nothing and every model is reported as having no divergences, indistinguishable from a model that genuinely never diverged.
- `loo_compare` printed the high Pareto-k count but never judged it. VG11's first LOO put 8,815 of 18,522 observations above k = 0.7 with p_loo = 10,720 and the row still entered the table looking like any other. The degeneracy tracks subject random effects exactly — 0% for every model without them, 30–49% for every model with them — so the notice fires past a fifth and points at leave-one-subject-out.
- `compare_ds_td_re` labelled console rows with the age *requested* rather than the one read, so every run printed a "30 mo" row carrying 25-month numbers. The CSVs were never affected.

## Verification

**957 passed, 45 skipped.** Ruff, Prettier and CSpell clean. The marginalisation carries 29 tests including density agreement with PyMC to 1e-9 and fine-grid integration; the figures carry 23 including a `max_age_months` alignment test verified by deleting the guard and watching it fail. All 18 fits of record were checked field-for-field against the current definitions before the VG22 change, and the two VG22 fits it invalidates are named above.


frankbuckley:
> [!NOTE]

`main` is merged in (2944e4c) and the PR is now **MERGEABLE**. Two of the five conflicts were semantic rather than textual — they merged cleanly as text and clashed in meaning — so recording what the resolution decided, because a reviewer would otherwise have to reconstruct it.

**`src/vocab_growth/models/common_bivariate.py` — a clean text merge that was wrong.** The merge auto-restored `extra_trace_var_names = ("kappa_u_obs", "kappa_s_obs")` from `main`, naming two variables this branch stopped sampling in 6c054bb. Resolved by keeping `main`'s restructured cross-lag diagnostics — suppressed understood LOO, reordered pair plot — and dropping the `extra_trace_var_names` machinery along with the variables it names. The model graph still defines those deterministics; they are simply not in `var_names` any more, so nothing else changes.

**`docs/models/README.md` — the important one.** Every row appeared to conflict, which turned out to be Prettier padding the table to this branch's longer VG22 row. Only three rows differ in content — VG09, VG16 and VG20 — and **`main` is right in all three**. In particular VG16: `main` withdraws the `beta_lag` = **+0.203** headline pending #242, while this branch still quoted it. Resolving in favour of "ours" would have silently restored a withdrawn statistical result. Took `main`'s table wholesale and appended only the two new VG21/VG22 rows.

The remaining three were genuinely additive: both sides appended rows to the notes index (merged and re-sorted into the table's ascending-timestamp convention), both appended test blocks to `tests/test_gp_utils.py` (both kept), and both added a condition to the same `if` in `common_bivariate_re.py` — `main`'s cross-lag flag and this branch's Down-syndrome-population scope, both kept. The population test is the one that matters there: it keeps the `survey_vocab_max` request off the typically-developing frame, which is built from a Wordbank query that never produces the column.

## Fit validity after the merge, checked rather than assumed

`main`'s engine commit states it is fit-preserving. Verified against every stored manifest: **16 of 18 fits still validate** field-for-field against the merged definitions. The two that do not are both deliberate, and neither is a merge artefact:

| fit | invalidated by | why |
| --- | --- | --- |
| `VG22-…-anchored-factor` | this branch (73dad85) | the registered default moved from rank 2 to rank 3 |
| `VG16-…-crosslag` | `main` (91f07c2) | cross-lag sources reassigned per child-age wave, consistent with withdrawing its headline |

## One local-only test failure, not a CI one

Three `tests/test_reporting_age_policy.py` cases fail locally for VG14/VG15, and they are stale stored output meeting `main`'s new sign-ratio cap rule from 565a769 (#238) — the artefacts on this machine were written 2026-08-22, before it. This branch touches neither `reporting_ages.py` nor that test, and the merged emitters do trim with `max_age_for_sign_ratio` correctly, so the code is right and the files are old.

Worth one caveat for whoever clears them: `regenerate_plots.py` is **not** enough. Regenerating VG14 rewrote `modality_trajectories.csv` and fixed that case, but left `posterior_summary_r.csv` and `posterior_summary_p_any.csv` at their 2026-08-22 contents, still running to 84 against a cap of 72 — exactly the two tables `reporting_ages.max_age_for_sign_ratio` names as having been trimmed with the wrong cap. Those are fit-stage artefacts, so clearing them needs a refit.

None of this reaches CI: the affected cases are `skipif(not FITTED)` and parametrised over models found on disk, and CI has no `output/models/`. Which is also why the defect could land with its test green.

## Still outstanding, and not blocking review

VG22's canonical rank-3 fit and its rank-2 variant fit, both invalidated by the rank move above. Neither can be produced by relabelling the existing output: `validate_fit_output` compares the whole recorded definition, and the sensitivity machinery writes its suffix into both `config_name` and `banner`, so the stored rank-3 fit differs from the new canonical in exactly those two fields despite an identical graph. Same precedent VG21 sits under.

PR #334, opened 2026-09-11
Cut the slow test set's cost where it actually was (#331)

### LINKED ISSUE #331: Cut the slow test set's remaining cost: it is CPU-bound after the distribution fix
> [!NOTE]

## Where this starts

`notes/202609101310-slow-test-distribution.md` profiled the `tests-slow` job and found its 10 m 20 s was the cost of a single file: `--dist loadfile` pinned `test_prior_table_coverage.py`, 106 tests and 45% of the set's 1,320 s of CPU, to one worker. Switching that job to `--dist loadgroup` with `xdist_group` marks on the three modules that share an expensive fixture fixed the distribution, and moving the priors-table exemption check to the fast set removed 21 tests that were never doing slow work.

That leaves the job **balanced rather than critical-path bound**: measured 592 s → 207 s at CI's four workers, and the floor is now total CPU divided by workers rather than the cost of one file. Every remaining lever is therefore either *less total work* or *more workers*. Four candidates, measured or costed, roughly in order of value.

## 1. Stop `test_prior_table_coverage` preparing real data (biggest single cut)

It is 591.7 s of the 1,320 s. Each of its 106 tests runs the engine's real `prepare` → `priors` → `build`, so 106 full DuckDB loads and 106 full-size graph builds — to read back a set of **parameter names** from `diagnostics_var_names`.

`test_graph_equivalence.py` gets the same names from `support.synthetic_graphs.build_synthetic_model`, which deliberately skips the preparation stage and supplies a small synthetic frame instead. If the reported name set is identical under a synthetic frame, this file's cost falls by most of that 591.7 s and the whole slow set roughly halves.

What has to be established first, and it is mechanically checkable in one run: that for all 21 registered models **and all 85 registered variants** the reported names are the same under a synthetic frame as under the real one. The risk is a variant whose parameter set depends on the *data* rather than the definition — the loader variants (`us01-implausible-reinstated` and friends) change which rows are admitted, and a parameter whose existence follows from the number of studies or children in the frame would differ. Write the equivalence check as a throwaway script, run it over all 106, and only then change the tests.

## 2. `FAST_COMPILE` where a function is compiled once and evaluated once

The set's cost is compilation, not sampling — the heaviest single test samples 8 draws in 1 chain and names nutpie's graph compilation as its cost. `support.synthetic_graphs.fixed_point_logp` calls `model.compile_logp()` and evaluates the result exactly once, which is the pattern PyTensor's `FAST_COMPILE` mode exists for: it skips the rewrite passes that dominate compile time and pays for it in execution speed that this caller never needs.

Two things to check before adopting it: that `tests/support/graph_baseline.json`, which records log-probabilities to a 1e-9 relative tolerance, still matches — it was generated under the default mode, and while skipping rewrites should move the value less than the ~4e-16 reassociation effect the test's own comment describes, that is exactly the assumption a guard test should not be asked to take on trust; and that the mode is set narrowly at the call site rather than globally, so nothing that *does* need a fast function silently gets a slow one.

## 3. Cache the PyTensor compiledir in CI

The jobs restore uv's package cache but nothing else, so every run compiles from cold — and compilation is the whole cost. Worth measuring what a warm compiledir saves (compare a run with `PYTENSOR_FLAGS=base_compiledir=<fresh temp>` against the default) before adding the machinery.

If it is adopted, key the cache strictly on the Python version and the pytensor, numba and numpy versions, with **no loose `restore-keys`**. A stale numba function cache is not a hypothetical failure mode here: on 2026-09-10 it produced a `CPUDispatcher` typing error that survived reruns and per-module retries until the offending `string.deepcopy-*` entries were deleted by hand, and on 2026-09-09 the cold-cache variant of the same race failed one fit test transiently. See `memory` notes and `notes/202608281312-ci-critical-path.md`.

Related mitigation, worth doing whether or not the cache lands: give each xdist worker its own `base_compiledir`. The distribution fix means 106 graph builds that used to compile serially on one worker now compile in parallel on four, which increases concurrent writes to the shared compiledir — the condition behind that race. Per-worker directories trade cache sharing for isolation, which costs nothing on a cold CI cache.

## 4. More workers: shard the job, or a larger runner

Since the job is now CPU-bound at four workers, wall clock falls roughly linearly with worker count: 207 s on four and 118 s on thirty-two, measured. Either a larger runner, if the organisation has them, or a matrix over test-file groups. Sharding costs a repeated `uv sync` and `prepare_data` per shard (~45 s) and more runner minutes, so it is the lever to pull *after* 1 and 2 have removed the work that does not need doing.

## An open question the profiling raised

`test_graph_equivalence.py`'s `built` fixture eagerly builds all 21 registered graphs for 107 cheap tests. Grouped onto one worker that module takes 40.5 s; ungrouped across four workers — each rebuilding all 21 — it takes **21.2 s**, because the builds then run in parallel. Grouping still looks right for CI, where the full set is CPU-bound and duplicating build passes raises total CPU, but the trade-off is worth revisiting if 1 and 2 land and the set stops being CPU-bound. A lazy per-model cache in that fixture would make the question moot: each worker would then build only the models its own tests ask for.

Also unresolved: whether `tests-fast` should move to `--dist loadgroup` too. It is 3 m 31 s and was not profiled, so this is a question rather than a proposal.

## Done when

- The equivalence check in 1 has been run over all 21 models and 85 variants, and either the tests are switched or the reason they cannot be is recorded;
- 2 and 3 are each either adopted with a measured saving or rejected with the measurement that rejected them;
- `notes/202609101310-slow-test-distribution.md` §5 is updated with what happened, and `ci.yml`'s comments and the three agent-instruction copies still describe the job that exists.

frankbuckley:
> [!NOTE]

Closes #331. Full record in [`notes/202609111158-slow-test-cost-was-not-data-preparation.md`](notes/202609111158-slow-test-cost-was-not-data-preparation.md).

## The headline: the issue's first item was aimed at the wrong thing, and measuring it found the right one

#331 item 1 attributed `test_prior_table_coverage.py`'s 591.7 s — 45% of the slow set's CPU — to its 106 real data preparations, and predicted that a synthetic frame would roughly halve the whole set. Switching the file measured **6 m 01.8 s → 5 m 45.1 s: 4.6%**. Timing the engine stages directly shows why: `prepare`, `priors` and `build` together cost **6.6 s across all twenty-one models**, because `tests/conftest.py` already silences the two expensive parts of preparation (the prior figures and the `describe_all` pass).

**The cost was `diagnostics_var_names`.** `Variable.eval()` compiles a PyTensor function for its own graph, and the function asked every unobserved RV for its size one at a time — 30 to 98 compilations per model, about 60 ms each:

|                                        | total, 21 models |
| -------------------------------------- | ---------------: |
| build every graph                      |            6.4 s |
| ask each graph for its parameter names |       **75.8 s** |

`_element_counts` takes the static type shape where there is one (every scalar prior has one) and evaluates the remainder **together** in a single compiled function. Verified **byte-identical — both returned lists and their order — across all 21 registered models and all 85 registered variants, 106 of 106**, at 416.9 s against 97.7 s. This is production code, so every fit pays it too; at six seconds a fit, that is why it was never noticed there.

## What each of the issue's four items came to

| item | outcome |
| --- | --- |
| 1. Synthetic frame for `test_prior_table_coverage` | **Adopted, for a different reason than proposed** — see below |
| 2. `FAST_COMPILE` | **Rejected by measurement**: fails outright on 15 of 21 models |
| 3. Cache the CI compiledir | **Adopted, and the largest of the four**: 6 m 41 s cold vs 4 m 07 s warm |
| 4. Shard the job / larger runner | **Not attempted** — the remaining cost is real sampling |

### 1. The synthetic frame is kept on the contract argument, not the cost one

At 4.6% it would not be worth the risk on cost alone. It is kept because a priors table is built from a model's **definition**, so checking it against a data-dependent parameter set imports a dependency the contract does not have — the argument `test_graph_equivalence` has always made for itself. The file also no longer needs the prepared DuckDB, so it no longer skips without one.

The issue named the risk precisely ("a parameter whose existence follows from the *data*"), so it was checked over all 106 cases rather than argued:

- **In no case does the synthetic frame report fewer parameters than the real one.** The check can only get stricter this way.
- **One case differs, in the safe direction.** `vg15/dse-native-only` leaves exactly one psi-informed study in the real pool, so its documented single-study branch drops `tau_psi` and `z_psi`; the synthetic frame's four cross-tab studies keep them.
- **No free variable crosses the `size <= 2` boundary** separating the summary set from the gate set.
- **One case could not build synthetically**: `vg16/lag-same-form` reads `survey_vocab_max`. Added to the frame, constant across it — the frame's children span two studies by construction, so a per-study ceiling would drop every lag and take `beta_lag` out of the graph. It moved no entry in `support/graph_baseline.json`.

### 2. `FAST_COMPILE` fails on fifteen of twenty-one models

Where it works it is exact (2.1e-14 relative against the committed baseline, against a 1e-9 tolerance) and roughly halves the compile. Elsewhere it raises `NotImplementedError: Elemwise.perform (Python mode) does not support more than 32 operands` — `FAST_COMPILE` is `Mode("py", "fast_compile")`, and `Mode(linker="cvm", optimizer="fast_compile")` was tried as the obvious escape and fails on the same fifteen. Even a working version would save ~16 s of the 31.9 s the registry spends compiling.

### 3. The compiledir cache, and why the key is strict

Measured before any machinery: the slow set at four workers is **6 m 41 s cold against 4 m 07 s warm**, and CI restores uv's package cache and nothing else, so every run has been the cold number. The key hashes `.python-version` and `uv.lock` and carries **no `restore-keys`** — `uv.lock` pins every package, so any version move gives a new key and a cold, correct compiledir. That is deliberate: the 2026-09-10 `CPUDispatcher` typing error that survived reruns until stale numba entries were deleted by hand is exactly the cache a loose prefix key would restore.

**The per-worker `base_compiledir` mitigation is not adopted**, and the cache is why: it was costed as free "on a cold CI cache", but the cache is no longer cold and per-worker directories would split it.

## Also here

`test_graph_equivalence`'s `built` fixture builds lazily instead of eagerly constructing all twenty-one graphs, which closes the issue's open question about its `xdist_group`: a worker now builds only the models its own tests ask for whichever way tests are distributed. The mark stays, because the job's floor is total CPU.

The issue's other open question — whether `tests-fast` should move to `--dist loadgroup` — is answered "not without marks first": `test_trend_gp_consolidation.py` carries a session-scoped model cache whose own docstring records the `loadfile` dependency.

## Measured result

All on one machine, one day, same conditions.

| Measurement                                  |     Before |          After | Ratio |
| -------------------------------------------- | ---------: | -------------: | ----: |
| `test_prior_table_coverage.py`, serial       | 6 m 01.8 s | **1 m 29.6 s** |  4.0x |
| `test_prior_table_coverage.py`, 32 workers   |     39.8 s |     **21.9 s** |  1.8x |
| Slow set, 4 workers (CI's count)             | 5 m 48.9 s | **4 m 07.3 s** |  1.4x |
| Slow set, 32 workers                         | 2 m 56.5 s |     2 m 36.7 s |  1.1x |

That file is no longer the slow set's largest cost and has left its slowest-fifteen entirely; the top costs are now `test_observation_deterministics`'s module fixture (a real fit) and `test_bivariate_re_holdout_mask`'s sampling tests. Those are real work, so the next lever is more workers rather than less — item 4, untouched.

## One cost to note

`common.py` is under the hashed package, so this edit moves the executable-code signature for the `sync`/`publish` paths. **It costs nothing extra today**: `6860e21` (the role declarations, 2026-09-09) already restaled all twenty-one fits pending the next refit window, and this matches the standing plan that changes land before that window.

## Verification

- Full suite `-m "slow or not slow"` green; `ruff`, `mypy`, `npm run spellcheck` and `npm run format:check` all clean.
- `diagnostics_var_names` old vs new: 106/106 byte-identical.
- Synthetic vs real reported names: 106/106 checked, 0 cases where the synthetic frame reports fewer.
- `graph_baseline.json` unmoved by the `survey_vocab_max` addition (all 107 `test_graph_equivalence` tests pass).


frankbuckley:
> [!NOTE]

## The CI figures, now measured rather than extrapolated — and a correction to #332's note

The table in the description was local, on a 32-core workstation at `-n 4`. Measuring it on CI changed both the number and one claim in the record, so here it is properly. All from the `tests-slow` job's **pytest step** (job totals in brackets, which include a cold `uv sync`):

| Run                                                      |      pytest step |
| -------------------------------------------------------- | ---------------: |
| main `789c213`, before `--dist loadgroup`                | 10 m 23 s (10 m 35 s) |
| main `82cd646`, after it                                 |  9 m 33 s (9 m 48 s) |
| this branch, first run — cold compiledir, cache populated |  7 m 25 s (10 m 02 s) |
| this branch, rerun — **warm compiledir**                 | **4 m 16 s** (4 m 30 s) |

**9 m 33 s → 4 m 16 s, a 55% cut**, split 9 m 33 s → 7 m 25 s for the code change and 7 m 25 s → 4 m 16 s for the compiledir cache. The cache restored 24 MB and its post-step correctly declined to re-save on a hit. `tests-fast` is unchanged: 1 m 32 s → 1 m 38 s, which is a cold cache on its own first run.

So **the cache is the larger half of this change**, which was not obvious before measuring — #331 listed it third.

### The correction

[`notes/202609101310-slow-test-distribution.md`](notes/202609101310-slow-test-distribution.md) §3 reports the `loadgroup` change as **"592 s → 207 s on four workers"**, noting "the 592 s from CI and the rest locally". Those two halves are not comparable, and pairing them in one row reads as a CI result. **On CI that change was 10 m 23 s → 9 m 33 s: 8%, not 2.9x.** An `-n 4` run on a 32-core workstation is not an `ubuntu-26.04-arm` run.

The rebalancing was still worth having and is what made everything here visible — it removed the ceiling that meant the job could not finish faster than one file, without which the `diagnostics_var_names` saving would have been hidden behind that file's serial wall. But the magnitude in the record was wrong, so §5 of that note and the three agent-instruction copies now say what CI measured, and this PR's own note labels every local figure as local.

`e0c7fe4` carries the correction; nothing in the code changed.

frankbuckley:
> [!NOTE]

A second warm run landed while the docs commit went through, and it is slower than the first — **5 m 16 s against 4 m 16 s** for `tests-slow`'s pytest step. Both are warm-cache runs of this branch, so the spread is the hosted runner's rather than the change's. Recording both rather than the flattering one: against main's 9 m 33 s that is **45–55%**, not a single 55%.

The same run shows the cache reaching the other two jobs, which was expected but not previously observed: `tests-fast` 1 m 32 s → **1 m 07 s** on the pytest step, and `model-fit` 2 m 31 s → **1 m 56 s** as whole jobs. Neither is touched by this change except through the compiledir.

All six checks pass on `e0c7fe4`; the note now carries both figures.

Issue #297, opened 2026-09-04
Model proposal: VG25 — sign→speech cross-lag on VG15, deferred on #242


frankbuckley:
> [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Fable 5.1).

## Proposal, and why it is deferred

A new model, **VG25**: VG15 with a **sign → speech cross-lag**, the joint-engine analogue of VG16's understood → `q` term. A child's prior-wave signed share of comprehension, relative to a baseline, shifts the logit of their current production ratio `q` through one coefficient, `beta_sign_lag`. VG15 is nested at `beta_sign_lag = 0`.

**Deferred on #242.** Every open item on that issue applies to a second lag model unchanged: the administration-wave definition across forms (the complete-wave grouping has landed in `cross_lag.py`; what a wave is when a child has rows on two forms at one age has not been decided), the LOO leakage (leaving a signed observation out does not remove it from the later spoken predictors it feeds), wave-sequential recovery, and subject-marginal predictions that silently condition on no lag history. Building VG25 before those are settled on VG16 doubles the debt rather than paying it. The companion proposal #296 (VG24, the correlated subject block on VG15) is the between-child estimate that should come first regardless, by the argument in `notes/202608151140-cross-lag-not-for-models-of-record.md` §4. This issue records the design so it is ready when #242 closes, and so the decision to defer is on the record rather than implied.

## The question and the estimand problem

Asked on 2026-09-04: can we model properly whether early signing predicts later speech? `notes/202608160930-early-signing-and-later-speech.md` found, among 147 children and holding earlier speech and comprehension standing fixed, that the signed share of comprehension predicts later spoken vocabulary at +0.19 SD per SD, 89% ETI [0.03, 0.36], stable under study fixed effects and single-study deletion. The binary signer contrast is not identifiable, because signing is very nearly a study-level property. That note is a two-wave residual regression, not a model, and the report callout in `docs/comparison/index.qmd` quotes it as such.

A cross-lag is the obvious model, and the estimand it delivers depends on the baseline in a way that has to be chosen deliberately, not inherited:

[...]

Issue #225, opened 2026-08-16
Between-child scale recovered low in 9 of 9 replicates, with and without subject_variance_partition; mechanism unidentified


frankbuckley:
> [!NOTE]
> Drafted by an LLM-based AI tool (Claude Code/Opus 5).

> [!IMPORTANT]
> **Status 2026-09-07 — closed as superseded.** The `rep` runs (comment of 2026-09-06) changed what this issue is about: the `tau_subject` bias is real and reproducible (low by 5.32–8.05% in two clean VG12 `rep` replicates) but it sits inside a broad under-coverage of the whole VG12 posterior, VG11 misses in the **opposite** direction through the opposite channel, and both VG12 replicates carry BFMI below 0.3, so the intervals the coverage claim rests on are the ones the diagnostics call unreliable. The attribution to `subject_variance_partition` is not supported. Of the proposed work: 1 ran (VG11, `test`, one assessable replicate); 2 ran (VG12 at `rep`); 3 is registered (`vg12 free-scales`) but not run; 4 is withdrawn as premature; 5 is done (`3665815`). Everything still open is tracked elsewhere — the mechanism and the design options in [#229](https://github.com/dseinternational/vocabulary-growth/issues/229), the τ–κ reparameterisation that must precede any further VG12 recovery as [#289](https://github.com/dseinternational/vocabulary-growth/issues/289) task 4.6, and the re-run recovery as #289 task 4.8. The body below is left as written; it records the July–August position.

## Summary

Parameter recovery on VG12 shows `tau_subject` **systematically underestimated in every replicate**, by about 5.6%, with the truth outside its 89% interval all three times. The bias enters through `subject_variance_share` — the split the `subject_variance_partition` reparameterisation introduces — not through the total scatter budget, which recovers cleanly.

This matters beyond the model: `tau_subject` is what the DS-vs-TD between-child contrast reads on the typically-developing side, and `build_variance_partition`'s docstring states that "the DS/TD heterogeneity contrast that `tau_subject` feeds is unaffected". These runs are evidence that it is affected.

Affected models are **VG11** (`reference_proportion=0.0118`) and **VG12** (`0.1041`). No other model carries the partition. Only VG12 has been through recovery; VG11 is untested and structurally identical.

## Evidence

`scripts/fit_recovery.py vg12 --config test`, three replicates, truths drawn from VG12's own posterior. `z = (posterior_mean − truth) / posterior_sd`, so negative means the posterior sits below the truth.

[...]

PR #335, opened 2026-09-11
Support VG16 parameter recovery, and derive the simulation order (#297)

frankbuckley:
> [!NOTE]

Progresses [#297](https://github.com/dseinternational/vocabulary-growth/issues/297) on its one remaining dependency, and clears the VG16 half of [#289](https://github.com/dseinternational/vocabulary-growth/issues/289) task 3.9 / [#242](https://github.com/dseinternational/vocabulary-growth/issues/242) item 6. Full record in [`notes/202609111500`](notes/202609111500-vg16-recovery-and-the-blocker-that-was-not-there.md). No fit of record is touched.

## The headline: the blocker was a misdiagnosis

`recovery/spec.py` has recorded VG16 as unsupported since #163, because "the cross-lag predictor is a function of earlier-wave comprehension, so forward simulation must proceed wave by wave; **single-pass simulation would fit synthetic-lag data against real-lag truth**". #242 item 6, #289 task 3.9 and the deferral of #297 all rest on that sentence.

It is false for VG16, and building the fix is what showed it. The simulator has **always** drawn in rounds that rebuild the model in between, the lag is derived inside that build off `analysis_df["understood"]`, and VG16's lag reads a column drawn in **stage 0** while entering `y_s_obs` in **stage 1**. Forcing the single pass and comparing the predictor at each build against the one the finished frame implies:

| build                                     | `understood` differs from the finished frame | lag predictor differs |
| ----------------------------------------- | -------------------------------------------: | --------------------: |
| 0 (before any draw)                       |                               1,702 of 1,708 |             575 rows |
| 1 (the build whose draw consumes the lag) |         448 — all rows with no comprehension count |        **0 rows** |
| 2 (final coherence rebuild)               |                                          448 |        **0 rows** |

## What replaces the assertion

Not another flag — that is how the wrong answer got written down. `OutcomeDependentPredictor` declares what a predictor **reads** and what it **enters**; `single_pass_is_sound` derives the answer: one pass is sound exactly when the source is drawn strictly before every consumer.

**The wave loop is built, because the case is real — just not VG16's.** #297's VG25 reads the prior wave's `signed` share, and the joint engine draws `signed` in the *same* stage as the `spoken` it would shift. The derivation selects the wave loop there automatically. That is the piece #297 actually needed.

**The guard runs either way**, and it is the part that would have caught the original claim: after simulation, every row's predictor as drawn is compared against the predictor the finished synthetic frame implies, and a mismatch aborts. `tests/test_recovery_wave_sequential.py` shows it failing on arrays *and* on a real simulation with the ordering forced unsound — a guard never seen to fail is not evidence.

## VG16 parameter recovery, first ever

Three replicates, `test`, `--truth posterior`, 7–12 minutes each. **Only one is assessable** — the harness scores a replicate only on confirmed convergence, and two missed the R-hat ≤ 1.01 gate:

| replicate | converged | max R-hat | min ESS | coverage (89%) |
| --------- | --------- | --------: | ------: | -------------: |
| r01       | **yes**   |    1.0096 |     449 |          0.842 |
| r02       | no        |    1.0132 |     393 |              — |
| r03       | no        |    1.0105 |     537 |              — |

On r01, `beta_lag` **missed low by 29%**: truth 0.285, posterior median 0.201, 89% ETI [0.121, 0.284], the truth at the 94.8th percentile — just outside. All-quantity coverage 0.842 against nominal 0.89 is this engine's known shortfall (#225), not the lag's.

### One thing worth more replicates, stated as the impression it is

| replicate          | truth `beta_lag` | posterior median |
| ------------------ | ---------------: | ---------------: |
| r01 (assessed)     |            0.285 |            0.201 |
| r02 (not assessed) |        **0.045** |            0.195 |
| r03 (not assessed) |            0.267 |            0.242 |

Truths span 0.045–0.285; posterior medians span 0.195–0.242. Two of those three may not be quoted, so this is not a measurement — but it is the shape the VG16 review worried about, and r02 is by accident the `(beta ≈ 0)` cell #242 item 6 asks for. More replicates at a converging tier is now something that can simply be run.

## What this does not do

- **The three designed cells of #242 item 6 are still outstanding.** They need two things the harness does not have: a way to *set* a parameter in the truth draw rather than take what the posterior or prior offers, and the correlated-random-effects-plus-lag comparator, which is not a registered model.
- **VG25 is not registered.** #297's four design decisions and the registration itself are untouched, and neither waits on this.
- **A tier decision is now on the table**: two of three missing the gate at `test` says this check needs a higher tier, and that cost should be known before the designed cells are commissioned.

## Records corrected

The claim appeared in five places and all five now say what was measured: `recovery/spec.py`, `docs/runbooks/parameter-recovery.md`, `docs/models/README.md`'s VG16 row, the three agent-instruction copies (which also still said VG24 was unsupported — it has been a target since 2026-09-06), and `tests/test_recovery_spec.py`, whose "unsupported model explains itself" test used VG16 as its example.

## Verification

- Full suite `-m "slow or not slow"` green; `ruff`, `mypy`, spellcheck and `format:check` clean.
- 15 new tests, 2 of them `slow`, including the guard aborting a real simulation.
- VG16 simulation with the wave loop forced: 7 waves, guard passes. With it not forced: one pass, guard passes. With the ordering forced unsound: aborts.


frankbuckley:
> [!NOTE]

## Review pass: one real gap, and a regression check the change needed

`e207a20`.

### The gap: the wave loop shipped untested

VG16 turns out to be sound in one pass, so it takes the single-pass path — which means **both slow tests bypassed the wave loop entirely**. The loop that exists for VG25, the whole reason for the change, had no coverage. It had been run by hand (7 waves, guard passing) and nothing more.

There is now a test that forces it on VG16 and checks the invariant that matters: every row belongs to exactly one wave, so the two passes must write **the same likelihood-row counts** even though their drawn values differ. It also asserts from the provenance record that the run really went wave by wave (`waves: 7`, `single_pass_sound: false`) rather than falling through.

### The regression check the change needed and did not have

The single-pass path is the one every existing recovery target uses, so the first question is whether it moved. Simulating **VG15, VG07 and VG11** at a fixed seed, before and after the whole change, gives **byte-identical synthetic frames** (sha256 over the sorted-column CSV), identical `row_counts`, identical `simulated_columns`, identical truth draw. VG15 is the one that matters most — it exercises both compositions, both nested links and `totals_tracking_parent`.

The empty-rows skip I had added is now scoped to the wave loop, so that path is unchanged *by construction* as well as by observation.

### Two things now stated rather than left implicit

- **`_verify_coherence` holds only the last round's denominators.** Under the wave loop that verifies the last wave directly and the earlier waves only through the frame-level checks (cells summing to their total, counts inside their denominator, totals tracking their parent), which do cover every row. Sufficient here, because a wave's rows are final once drawn and their denominators cannot move afterwards — but weaker than the predictor guard's per-round check, and a future engine whose denominators depended on a *later* wave would need it strengthened. The code says so now.
- **The note showed "—" for the two unassessed replicates' coverage.** They were computed (0.752 and 0.820) and *refused*, which is not the same as absent. The table now shows them italicised with what the italics mean.

### The central claim, re-verified two ways

The load-bearing claim is that single-pass simulation was always sound for VG16. Beyond the measurement (0 of 1,708 rows differing at the build whose draw consumes the lag), it also holds structurally: `q_lag_term` appears in exactly one place in `common_bivariate_re.py` — added to `h_obs_re`, the production-ratio logit, at line 1003. `f_u_obs_re` has no lag term, so `y_u_obs`'s density does not depend on `beta_lag` at all, and drawing all comprehension in one pass is exactly what the factorisation says. Measurement and structure agree.

### Also confirmed

- **No model of record was touched.** `VG16-…/trace.nc` and its manifest are still 2026-09-08 01:27; the three recovery fits are in their own `-recovery-rNN` directories.
- Every quoted figure re-read from the CSVs: R-hats 1.009593 / 1.013174 / 1.010549, min ESS 449.2 / 393.2 / 537.3, 112 of 133 within the 89% interval, `beta_lag` truth 0.28460 against a posterior median of 0.20105 with ETI [0.12106, 0.28355] and truth quantile 0.947875 — the 89% ETI ends at the 94.5th percentile, so "just outside" is exact. Per-replicate fits 6m49–12m23, so "7–12 minutes" holds.
- Full suite `-m "slow or not slow"` green; ruff, mypy, spellcheck, format:check clean. 16 tests in the new module, 3 of them `slow`.

Issue #240, opened 2026-08-23
Address VG11-VG13 statistical specification, validation and reporting findings


frankbuckley:
> [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

## Summary

A code and statistical review of VG11, VG12 and VG13 found that their registered PyMC graphs are internally coherent: dimensions, indexing, likelihood parameterisation, study constraints, GP anchors and posterior-predictive construction passed inspection. The material findings concern the fixed-810 measurement model combined with age-invariant child effects, grouped prediction targets, duplicated TD administrations, prior-calibration scope, sampler and recovery evidence, the obsolete canonical VG13 age window, and several reporting-estimand errors.

The complete evidence, limitations, model-specific findings and remediation order are recorded in `notes/202608231537-vg11-vg12-vg13-statistical-review.md`. This issue tracks implementation and validation; the documentation PR deliberately does not close it.

## Immediate correctness and reporting

- [x] Deduplicate exact full source administrations before outcome projection, study filtering and child coding; record the removal count and add a regression test. _(Done 2026-08-23 in `9871036`: `SELECT DISTINCT *` over the complete source row before projection, collapsing 28 exact copies (22 into VG11's frame, 3 VG12's, 2 VG13's), with a fixture test for the source-identity semantics and a real-database test pinning 18,522 -> 18,500, 7,052 -> 7,049, 6,358 -> 6,356. Verified 2026-09-06: the 109 rows that still look identical in VG11's *projected* frame are same-child same-age pairs on **two different forms** (all 106 `Thal` ones), which the fix deliberately preserves; that is item 12's territory, not this one.)_
[...]

Issue #233, opened 2026-08-22
Correct VG19/VG20 statistical validation and reporting defects


frankbuckley:
> [!NOTE]
> Drafted by an LLM-based AI tool (OpenAI Codex/GPT-5).

## Summary

A read-only code and statistical audit found that the core VG19 and VG20 PyMC graphs are implemented coherently: their Cholesky algebra, subject indexing, age scaling, conditional paired likelihood routing, and unseen-child posterior prediction are correct. However, several validation checks do not exercise the new model structures, and some report claims are mathematically incorrect or stronger than the available evidence supports. These issues should be resolved before treating the model reports as statistically verified.

The audit used a freshly prepared temporary DuckDB analysis frame containing 1,431 rows and 767 children. Of 1,428 spoken observations, 973 use the nested conditional likelihood and 455 use the fallback likelihood. Both real-data model graphs built with finite initial log-probabilities; targeted tests and Ruff passed. The full test suite's only two failures were caused by the absent generated DuckDB, and those two tests passed when pointed at the prepared temporary database. Reporting-quality traces were not available, so published posterior values and convergence statistics were not independently recomputed.

## High-priority correctness findings

- [x] **Replace VG19's invalid zero-exclusion argument for random slopes.** `tau_subj_u_1` and `tau_subj_q_1` have continuous HalfNormal priors (`src/vocab_growth/models/gp_utils.py:171-172`). An equal-tailed interval must exclude exactly zero even when the posterior equals the prior, so the claims in `docs/models/vg19/index.qmd:75,184-200` and `notes/202608212000-vg19-gates-g2-g4-g5.md:24-35` cannot establish that either slope variance is nonzero. Use subject-level predictive comparison with VG10, a point-null/spike-and-slab formulation, or a simulation-calibrated practically negligible threshold. Do not interpret interval exclusion as rejection of `tau1 = 0`. _(Done, verified 2026-09-06. `vg19/index.qmd:142` now states the correct thing: tau_1 "carries a `HalfNormal` prior, so its posterior is supported on (0, inf) and an equal-tailed interval **must** exclude zero — even if the posterior is the prior and the data have said nothing". `:373` repeats it as a finding — "Neither rate spread is testable from its own interval" — and contrasts the offset-rate correlations, which do live on (-1, 1) with prior mass either side.)_

[...]

PR #339, opened 2026-09-11
Register VG25: the sign -> speech cross-lag on the joint engine (#297)

frankbuckley:
> [!NOTE]

Registers **VG25**: VG24 plus one coefficient. A child's signed share of comprehension at their previous administration wave, relative to their own persistent signing standing, shifts the logit of their current production ratio `q` through `beta_sign_lag ~ Normal(0, 0.5)`. VG24 is nested exactly at zero — checked on the observed log density, not by inspection.

Nothing is fitted and no fit of record was touched. Record: [`notes/202609111704`](https://github.com/dseinternational/vocabulary-growth/blob/main/notes/202609111704-vg25-registration-and-the-boundary-share.md).

## Why VG24 and not VG15

This is the interpretive argument, not a class-hierarchy detail. `rho_sign_q` is the persistent between-child sign–speech association; with it in the model, the lag has that part taken away and is left measuring the prospective, occasion-level quantity it is named for. Without it, [`notes/202608151140`](https://github.com/dseinternational/vocabulary-growth/blob/main/notes/202608151140-cross-lag-not-for-models-of-record.md) §3 applies — within-child deviations in this pool have no memory beyond the occasion, so the lag becomes a noisy proxy for the correlation. Deriving from VG24 also leaves VG24's own fingerprint untouched and makes it the exact nested null.

The project now has a home for each of the three questions that sound like one, and the comparison report says which is which:

| Question | Quantity | Where | Status |
| --- | --- | --- | --- |
| Do the children who sign more of what they understand also *say* more of it, as a persistent standing? | `rho_sign_q` | VG24 | 0.391 [0.231, 0.539] |
| At one moment, within a child's understood words, do saying and signing a word go together? | `psi` | VG15 / VG24 | 2.34 [1.89, 2.81] |
| Does a child who signed *more than usual for them* earlier go on to say a larger share later? | `beta_sign_lag` | **VG25** | registered, not yet fitted |

The descriptive +0.19 of [`notes/202608160930`](https://github.com/dseinternational/vocabulary-growth/blob/main/notes/202608160930-early-signing-and-later-speech.md) is closest to the third but is not it — that regression conditions on the child's speech at the *earlier* wave, which a `q` child intercept does not.

## The design decisions

Four were named on [#297](https://github.com/dseinternational/vocabulary-growth/issues/297). The first two were the study owner's call; measurement forced a fifth.

| | Registered | Why |
| --- | --- | --- |
| Which likelihoods the lag enters | spoken marginal **and** the cross-tab compositions | one scalar on a fixed covariate cannot do to `psi` what a free per-child offset did — and confining it costs most of the evidence |
| Baseline | `within` (VG16 registers `population`) | `rho_sign_q` is in the model, so the population baseline has nothing left of its own to measure |
| `nz_01` denominator | no lag source | the signed share of *production* is a different variable — declined on that argument, and **not** because it is free |
| Gap | unrestricted, with a registered 12-month arm | one coefficient across 2–15 months is the assumption [#242](https://github.com/dseinternational/vocabulary-growth/issues/242) asked to be checked |
| **Boundary shares** | **continuity, not VG16's clip** | see below |

### Support, measured on the 2026-09-11 frame

| design | supporting observations | children | gap median (IQR) | by study |
| --- | ---: | ---: | --- | --- |
| marginal **and** cells (registered) | 191 | 129 | 6 (4–11) | uk_07 52, ie_02 43, uk_02 41, uk_05 30, uk_04 25 |
| marginal only | 111 | 80 | 6 (3–7) | ie_02 43, uk_05 30, uk_04 25, uk_02 13 |

`es_01` contributes 185 rows carrying a signed share and **not one lag** — all 186 of its children are single-wave, verified directly.

**`nz_01` costs real support.** 28 of its 33 children have more than one wave, and admitting a produced denominator would take the coefficient to **269 observations over 157 children** — 78 added, at a median gap of 6 months, and the 28 children #297 anticipated. It is declined on the argument alone, which is the only way the rule survives the next `nz_01` follow-up.

**The caveat, stated rather than left in the code.** Under the `within` baseline the predictor itself contains the child's signing shift at the *prior* wave, so an estimated per-child quantity does reach the composition — through one scalar coefficient, on lagged rows only, rather than as a free offset per row. `sign-lag-population` is the arm in which nothing per-child reaches the cells at all.

### The fifth decision, which the issue did not anticipate

VG16's clip is the wrong treatment for a **ratio** predictor, and the difference is not cosmetic.

| | VG16's count lag | VG25's ratio lag |
| --- | ---: | ---: |
| supporting rows | 477 | 191 |
| at a logit boundary | 7 (1.5%) | **28 (14.7%)** |

The clip puts all 26 zero-share sources at `logit(1e-4) = -9.21` **whatever the wave measured** — their denominators run 2, 7, 7, 13, … 129, 406. Measured on the **source signed-share logit** (the predictor's observed input, before the latent baseline the fitted term subtracts — the part measurable without a fit):

| treatment | SD | range | boundary rows' share of its sum of squares |
| --- | ---: | --- | ---: |
| clip | 3.34 | [−9.21, +9.21] | **76.1%** |
| continuity | 1.79 | [−6.70, +4.89] | 45.8% |

Under the clip, 14.7% of the rows carry three quarters of the variation `beta_sign_lag` is estimated from. **Not an artefact of the raw scale**: residualising on the source wave's age and study — the closest stand-in for the model's baseline available before a fit — leaves 66.1% against 42.3%.

The **field** default stays `LAG_ZERO_CLIP`, so the two lags are configured identically and VG16 is untouched; `sign-lag-clip` measures the difference on a fit.

## The wave loop, on the model it was built for

[#335](https://github.com/dseinternational/vocabulary-growth/pull/335) built the wave-sequential simulator and noted that no registered model selected it. VG25 does, and nothing declares that: `outcome_dependent_predictor` says the lag reads `signed`; `single_pass_is_sound` finds that column drawn in the same stage as every node the lag enters, and selects the loop. Measured end to end — **~70 s at `dev`, seven waves, sixteen coherence checks passed** — and now a test.

## Seven sensitivity arms, each measured

Registered in the same change as the fields they need — a field with no variant using it is dead weight in a fingerprint.

| arm | obs | children | what moves |
| --- | ---: | ---: | --- |
| _(registered)_ | 191 | 129 | — |
| `sign-lag-clip` | 191 | 129 | the predictor's **values**, not its rows |
| `sign-lag-marginal-only` | 111 | 80 | drops `uk_07` entirely and 28 of `uk_02`'s 41 |
| `sign-lag-population` | 191 | 129 | the **baseline** subtracted |
| `sign-lag-uk07-marginal` | 191 | 129 | `uk_07`'s 52 rows move from the `cells` branch to the `marginal` one |
| `sign-lag-gap-12` | 168 | 112 | drops the 23 lags reaching back over a year |
| `beta-sign-tight` / `-wide` | 191 | 129 | the prior scale only |

The four that leave the support at 191 are not inert — they change the predictor, the baseline, the likelihood carrying it, or the prior. Stated because a reader checking an arm against its row count would otherwise conclude three did nothing.

## Also in this change

- **`cross_lag.py` refactored** so both lags share one wave walk, one source-selection rule and one gap ceiling. `same_form_only` deliberately does **not** carry over: the ratio divides two counts scored on the same form in the same administration, so form truncation largely cancels — a property of the predictor, so VG25 carries no such field rather than one that would always be inert.
- **A report template written from scratch**, not derived from VG24's — copying a parent's report is how VG21 inherited three withdrawn claims ([#240](https://github.com/dseinternational/vocabulary-growth/issues/240)). It computes its support from the fit's own audit artefact, reads the boundary treatment from the manifest rather than assuming the registered one (a variant fit renders this template), and reports the coefficient's prior-to-posterior contraction — because VG24's `+0.000` correlations showed that a headline pinned at zero looks exactly like a model that fitted and found nothing.
- **The comparison report's "does early signing predict later speech?" callout** now names all three quantities.

### My own pass, before opening

Worth listing, because two of these were wrong in a way that reads as fine:

1. **The page declared itself a model of record.** `render_reading_routes('record', …)` routes families and practitioners to its expectations tables. VG25 is `UNCLASSIFIED` and unfitted, so it now uses the `candidate` role, which sends them to VG24 and says nothing here is a reported number.
2. **"The `nz_01` decision costs nothing measurable" was false.** It costs 78 observations from 28 children. Corrected in the definition, the model page, the inventory and the note, and now stated as a cost accepted on principle.
3. **A silent NaN was one field away.** The continuity correction has no clip of its own, so a source with `signed > understood` gives a share above 1 and `log(1 - r)` returns NaN into the log density. It cannot arise today (the loader masks comprehension below the recorded production union; cross-tab cells sum to their own total — verified, 0 of 562 rows), but reinstating that mask for a sensitivity is one field away. Now clipped in the primitive, with a test, and a no-op on valid data.
4. **"76.1% of the predictor's sum of squares" was imprecise** — that is the *source logit's*, since the fitted predictor subtracts a latent baseline. Restated, and corroborated with the residualised figures above.

### Automated review, after (1b592f4)

Nine comments, all accepted — and the three GitHub had *suppressed* were the substantive half.

1. **The source-selection rule was not order-independent, though its docstring said it was.** The wave walk ranked candidates on the comprehension denominator and took `np.argmax`, which is sound only while the ranked quantity is the *only* thing read off the chosen row. True of VG16's count lag as first written; false of a ratio lag that also reads the numerator. On two forms tied at 100 understood and disagreeing at 20 and 60 signed, swapping the rows moved the predictor from **−1.386 to +0.406**. The same hole had been open in VG16 since [#242](https://github.com/dseinternational/vocabulary-growth/issues/242)'s `same_form_only`, which reads the source's form ceiling — there the swap flips whether the row has a lag at all. The rule is now *every quantity read off the selected row is a selection key*, keyed lexicographically so the invariant is structural. No source wave offers a choice today, and every array is bit-identical to the committed implementation across VG16's registered form, its two arms and all seven VG25 arms. Ten waves do carry two usable comprehension measurements.
2. **The report's support tables ignored `sign_lag_in_cells`** — a cross-tab row has a likelihood branch whether or not the lag reaches it, so rendering `sign-lag-marginal-only` would have reported the registered model's 191/129 in place of its own **111/80**, and every gap, boundary, study and branch figure with it. Both cells now read the field from the fit's manifest, as the boundary treatment already was.
3. **The pair plot could not show the ridge the report names** — [#233](https://github.com/dseinternational/vocabulary-growth/issues/233) again, on the joint engine. `beta_sign_lag` is eighteenth in build order and the cap keeps six, so the grid would have shown `psi`, `conc` and four understood-GP hyperparameters. Fixed by ordering, not filtering, and then generalised — see below.

The other six were stale counts and headings: the field count (**seven**, not six), "four design decisions" over five subsections, the clamp docstring and its test name (sixteen of twenty-two), a sensitivity comment that reintroduced the "predictor" imprecision fixed above, and VG25's absence from the fit-command and recovery-target lists in the three instruction copies.

## The pair-plot consolidation, and what it found in VG24

The joint engine led with a hard-coded `("psi", "conc")`; the bivariate engine had `pair_plot_priority`. They are now one implementation in `diagnostics_utils`, which also disposes of the duplicated reordering closure.

I had argued against doing this inside a registration, on the ground that it would reorder VG14's, VG15's and VG24's plots. **Two thirds of that was wrong**, and doing it is what showed it:

- **VG14** is on the trivariate engine, which installs no reordering at all. Never affected.
- **VG15** does not move — bare `("psi", "conc")` before and after.
- **VG24** does move, and it had the same defect *live*. Its report calls a near-±1 correlation "the failure mode to watch for in the energy and pair plots", and the grid showed `psi`, `conc` and four understood-GP hyperparameters — not one of its three child correlations. The narrow fix would have left that.

So the reporting change is to **one** fitted model, and it is a repair. Measured across all twenty-two registered models — twenty unchanged:

| | rendered before | rendered after |
| --- | --- | --- |
| VG15 | `psi conc p_slope_low_u p_slope_hi_u ell_unit_u eta_u` | unchanged |
| VG24 | `psi conc p_slope_low_u p_slope_hi_u ell_unit_u eta_u` | `psi conc rho_sign_q rho_u_sign rho_uq tau_subj_u` |
| VG25 | `psi conc beta_sign_lag rho_sign_q tau_subj_sign p_slope_low_u` | `… rho_u_sign` sixth |

Consolidating was **not** applying the bivariate rules to joint models. A naive merge gives VG24 the priority `('rho_uq', 'tau_subj_u', 'tau_subj_q', 'tau_u', 'tau_q')` — `rho_uq` is the understood–spoken correlation, and the one VG24 exists to estimate is `rho_sign_q`. That is #233's own mistake repeated. The correlated-child branch is modality-aware; the head is kept separate from the definition-driven part so VG15 does not pick up a trailing scale block it has no structure to justify.

The fail-closed half: three engines install no reordering, which is only safe while their models have nothing to order. A test asserts that for every registered model on them, so registering a model with a distinguishing structure on an unwired engine fails and names the engine.

The figures change when VG24 is next fitted; this is a code change.

## Scope of change

The seven new fields live on `JointCrossLagModelDefinition`, a subclass no other model instantiates, so **no existing definition's serialisation moves** and no prepared-frame hash changes. `tests/support/graph_baseline.json` gained 315 lines and changed none — the evidence that the registration was graph-inert for the other twenty-one models.

It does move the **executable-code signature**, as any change inside `src/vocab_growth/` does, so existing fits need the refit [#289](https://github.com/dseinternational/vocabulary-growth/issues/289) already schedules before they can be published or resumed.

## What is deliberately left open

**#297 stays open.** This settles the design and registers the model; its adoption gates are unmet.

1. **No fit.** The role is deliberately `UNCLASSIFIED`, which fails closed — full publication strictness, and a place in the default refit scope (now 8 of 22). Whether VG25 becomes a model of record, a development step, or neither is a study-owner decision with [#190](https://github.com/dseinternational/vocabulary-growth/issues/190)'s other scope questions (#297 check 7). **Registering a model makes its output subject to publication validation immediately**, so delete any provisional VG25 output before syncing.
2. **Gate 4's three recovery cells.** VG25 is a registered recovery target and the loop is selected automatically, so `(beta ≠ 0, rho ≠ 0)` is runnable from the prior today. The other two need a way to *set* a parameter in the truth draw, which the harness does not have. Expect to need a tier above `test`: VG16's recovery returned `beta_lag` 29% low on its one assessable replicate, and two of three missed the R-hat gate.
3. **No leave-one-study-out arm** (#297 check 5). `exclude_studies` is a `BivariateModelDefinition` field; adding it to `JointModelDefinition` would restale every VG15 and VG24 fit — a change to make deliberately, with the refit it costs.
4. **LOO leaks across the lag**, as it does for VG16. `scripts/wave_forward_score.py` is the answer and has not been run here.

**Expect a small coefficient with an interval near zero.** The descriptive note's 89% interval only just cleared zero at n = 147, and this lag's support is 191 rows against VG16's 473. The deliverable is the estimate with its uncertainty in interpretable units, not a sign.

## Verification

- `uv run pytest -m "slow or not slow"` — full suite green at **2515 passed, 53 skipped** (+60 new tests).
- ruff, mypy, Prettier, CSpell clean.
- Graph baseline regenerated for `vg25` alone, so the diff can only add.
- Every quantitative claim above re-derived from the frame in a second, independent pass.
- All 20 Python cells in the report template compile; all 9 `@sec-` cross-references resolve.


frankbuckley:
> [!NOTE]

All nine accepted, and all three suppressed ones too — those turned out to be the substantive half of the review. Fixed in 1b592f4.

## The three that were real defects

**1. The source-selection rule was not order-independent, though its docstring said it was** (`cross_lag.py`). Right, and demonstrable. The walk ranked candidate sources on the comprehension denominator and took `np.argmax`, which is sound only while the ranked quantity is the *only* thing read off the selected row. That held for VG16's count lag as first written; it does not hold for a ratio lag that also reads the numerator. On a wave with two forms tied at 100 understood and disagreeing at 20 and 60 signed, swapping the two rows moved the predictor from **−1.386 to +0.406** — a sign change.

It reaches further than the comment said: `same_form_only` (#242) reads the source's **form ceiling** off the chosen row, so VG16 has had the same hole since that flag landed. There the swap flips whether the row has a lag at all. Both are now keyed on the stated rule — *every quantity read off the selected row is a selection key* — with the larger signed count and the larger ceiling winning, which is the least-truncated principle the primary rank already uses.

Rather than a tie-breaker on top of the old selection, the keys are lexicographic, so the invariant is structural. Nothing on the current frame exercises it: **no wave that serves as a source offers a choice, for either lag**, and I checked every array is bit-identical against the committed implementation across VG16's registered form, its same-form and gap arms, and all seven VG25 arms. Ten waves *do* carry two usable comprehension measurements, so one further administration for any of those children would have made it live.

**2. `sign_lag_in_cells` ignored in the report's support filter** (suppressed, `index.qmd:170` and `:537`). Confirmed by measurement: the old filter returns 191 observations / 129 children for `sign-lag-marginal-only`, which actually has **111 / 80**. Both cells now read the field from the fit's own manifest, as the boundary treatment already was. The giveaway was that the caption said the compositions count "only under `sign_lag_in_cells`" while the code counted them regardless.

**3. The pair plot could not show the ridge the report names** (suppressed, `index.qmd:397`). Confirmed: `beta_sign_lag` is **eighteenth** in VG25's build order and the cap keeps six, so the grid would have been `psi`, `conc` and four understood-GP hyperparameters. This is #233 again on the other engine, so it is fixed the same way — by ordering, not filtering, so nothing is hidden. The plot now shows `psi`, `conc`, `beta_sign_lag`, `rho_sign_q`, `tau_subj_sign`.

One deliberate narrowing: I did **not** route the joint engine through the bivariate `pair_plot_priority`, which would be the general fix. It would also reorder VG14's, VG15's and VG24's pair plots, and a reporting change for three fitted models should not ride along inside a registration. Only the lag's own block is prepended, so every joint model without the lag gets a byte-identical plot — pinned by a test. The cap regression test you asked for is in `tests/test_pair_plot_limits.py`, in the same idiom as #233's.

## The six stale counts and headings

All correct, all fixed: the field count (**seven**, not six — in both the dataclass docstring and the note), the report's "four design decisions" over five subsections, the test name (`..._by_sixteen_of_the_twenty_two`), the "predictor" wording in `test_sensitivity_overrides.py` that reintroduced exactly the imprecision this branch set out to fix, and VG25's absence from the fit-command and recovery-target lists in the three instruction copies.

The clamp-field one went one level further than the comment: `clamp_targets`' own docstring — the one that warns the count goes stale and says so at length — still read "fifteen of the twenty-one". It is now sixteen of twenty-two, and says the test asserts the *list* and the number is only ever a reading aid.

## Verification

Full suite `-m "slow or not slow"` green at **2513 passed, 53 skipped** (+7 tests: four for the tie-break, three for the cap). ruff, mypy, Prettier and CSpell clean. All 20 Python cells in the template compile and all 9 `@sec-` references resolve.

frankbuckley:
> [!NOTE]

Follow-up to my last comment: the pair-plot consolidation is now done in 733c988, and it **corrects something I said there**.

I had argued for the narrow fix on the ground that routing the joint engine through the bivariate `pair_plot_priority` "would also reorder VG14's, VG15's and VG24's plots". Two thirds of that was wrong:

- **VG14** is on the trivariate engine, which installs no pair-plot reordering at all. It was never affected by the joint engine's ordering, so it cannot be affected by consolidating it.
- **VG15** does not move. Its priority is the bare `("psi", "conc")` before and after, because it carries no distinguishing child structure.
- **VG24** does move — and it turns out it had the same #233 defect **live**, which the narrow fix would have left in place. Its report tells the reader a near-±1 correlation "is the failure mode to watch for in the energy and pair plots", and the grid showed `psi`, `conc` and four understood-GP hyperparameters, not one of its three child correlations.

So the reporting change is to one fitted model rather than three, and it is a repair rather than a restyle. Measured across all twenty-two registered models, twenty are unchanged:

| | rendered before | rendered after |
| --- | --- | --- |
| VG15 | `psi conc p_slope_low_u p_slope_hi_u ell_unit_u eta_u` | unchanged |
| VG24 | `psi conc p_slope_low_u p_slope_hi_u ell_unit_u eta_u` | `psi conc rho_sign_q rho_u_sign rho_uq tau_subj_u` |
| VG25 | `psi conc beta_sign_lag rho_sign_q tau_subj_sign p_slope_low_u` | `… rho_u_sign` in the sixth slot |

Two things worth flagging about the merge itself, because neither was mechanical.

**It is not a matter of applying the bivariate rules to joint models.** A naive merge gives VG24 the priority `('rho_uq', 'tau_subj_u', 'tau_subj_q', 'tau_u', 'tau_q')`. `rho_uq` is the understood–spoken correlation; the one VG24 was registered to estimate is `rho_sign_q`. That is #233's own mistake repeated — prioritising the parameter the *other* model was added for. The correlated-child branch is modality-aware, and on the joint engine it names all three correlations with `rho_sign_q` first.

**The head has to stay separate from the definition-driven part.** On the joint engine `psi` and `conc` lead unconditionally, so a list built by appending to them is never empty — and VG15 would have picked up the trailing scale block that is meant to mark a model with a structure worth prioritising. Pinned by a test.

Also fail-closed now: three engines install no reordering, which is only safe while their models have nothing to order. A test asserts an empty priority for every registered model on them, so registering a model with a distinguishing structure on an unwired engine fails and names the engine rather than silently dropping the intent.

The figures themselves change when VG24 is next fitted — this is a code change, and the cached figures are the previous fit's.

Full suite green at **2515 passed, 53 skipped**; ruff, mypy, Prettier and CSpell clean.

